PR #2, opened 2023-10-27
Add Zephyr driver api

jukkar:
This is very much WIP.

### DIFF
--- a/src/drivers/driver.h
+++ b/src/drivers/driver.h
@@ -2328,6 +2328,14 @@ struct wpa_signal_info {
 	int center_frq2;
 };
 
+
+struct wpa_conn_info {
+	unsigned short beacon_interval;
+	unsigned char dtim_period;
+	bool twt_capable;
+};
+
+
 /**
  * struct wpa_channel_info - Information about the current channel
  * @frequency: Center frequency of the primary 20 MHz channel
@@ -3785,6 +3793,13 @@ struct wpa_driver_ops {
 	 * @signal_info: Connection info structure
 	 */
 	int (*signal_poll)(void *priv, struct wpa_signal_info *signal_info);
+
+	/**
+	 * beacon_poll - Get beacon info of current SSID
+	 * @priv: Private driver interface data
+	 * @beacon_info: Beacon info structure
+	 */
+	int (*get_conn_info)(void *priv, struct wpa_conn_info *conn_info);
 
 	/**
 	 * channel_info - Get parameters of the current operating channel
@@ -6167,5 +6182,9 @@ extern const struct wpa_driver_ops wpa_driver_atheros_ops;
 #ifdef CONFIG_DRIVER_NONE
 extern const struct wpa_driver_ops wpa_driver_none_ops; /* driver_none.c */
 #endif /* CONFIG_DRIVER_NONE */
+#ifdef __ZEPHYR__
+#define CONFIG_DRIVER_ZEPHYR
+extern const struct wpa_driver_ops wpa_driver_zep_ops; /* driver_zephyr.c */
+#endif /* __ZEPHYR__ */
 
 #endif /* DRIVER_H */
--- a/src/utils/os_zephyr.c
+++ b/src/utils/os_zephyr.c
@@ -5,7 +5,8 @@
  * See README for more details.
  */
 
-#include <sys/time.h>
+#include <zephyr/posix/time.h>
+#include <zephyr/posix/sys/time.h>
 
 /* The clock_gettime() would be found in <zephyr/posix/time.h> but
  * that will cause conflict with picolib definition.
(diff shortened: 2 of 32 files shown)


src/drivers/driver.h
GannicusGuo, line 1: why change open source wpa_driver_ops structure to add this get_conn_info?

wpa_supplicant/bss.c
GannicusGuo, line 1: why change open source bss.c add this wpa_bss_check_pending_connect?
will this change merge to https://w1.fi/cgit/hostap/?

wpa_supplicant/wpa_supplicant_i.h
GannicusGuo, line 1: why change open source structure wpa_supplicant to add "connection_a, connection_b, connection_g "

wpa_supplicant/bss.c
krish2718: This was submitted [upstream](https://patchwork.ozlabs.org/project/hostap/patch/20230420192821.376828-1-krishna.t@nordicsemi.no/) but not yet merged.

wpa_supplicant/wpa_supplicant_i.h
krish2718: This is in the list of things to submit upstream but not done yet.

src/drivers/driver.h
krish2718: This is in the list of things to submit upstream but not done yet.

wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -0,0 +1,3695 @@
+/*
```
krish2718, line 1: FYI, This change to split generic commands in wpa_cli.c in to wpa_cli_cmds.c also was planned to be submitted upstream, but haven't gotten around to that.

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -37,7 +37,7 @@ extern struct wpa_global *global;
 
 static void wpa_cli_msg_cb(char *msg, size_t len)
 {
-	wpa_printf(MSG_INFO, "%s\n", msg);
```
krish2718: can this commit be squashed in to the `fix newline` commit?

dleach02:
FYI, if we are doing commits here and submitting changes to the hostap upstream, we really should consider adopting the notation Nordic uses for engaging with upstream projects... e.g. [zephyr fromlist] 

It will make it clear to the reviewers what is going on.

@carlescufi, thoughts?

src/drivers/driver.h
gannicus-gj: Hi @krish2718 , is it that hostap (upstream) needs to adopt the "list of things to submit" first,
before merging this "hostap/pull/2" to zephyr?  or there is no such dependency?

gannicus-gj:
Hi @krish2718 , is it that hostap (upstream) needs to adopt the "list of things to submit" first,
before merging this "hostap/pull/2" to zephyr? or there is no such dependency?

jukkar:
> FYI, if we are doing commits here and submitting changes to the hostap upstream, we really should consider adopting the notation Nordic uses for engaging with upstream projects... e.g. [zephyr fromlist]

That is fine by me, we can certainly do that if needed. Although that means that we would need to do rebases in this repo, that is fine too but we do not seem to do that in other zephyr repos.

dleach02:
I'm pointing this out because of the number of comments/responses around changes that have been submitted to the upstream. Most of the other modules we have in Zephyr usually don't have too many changes like this and I'm not sure if hostap will be different in this way.

Again, just a suggestion. I like the Nordic tagging/workflow as it does clearly identify the origin of a commit.

krish2718:
> Again, just a suggestion. I like the Nordic tagging/workflow as it does clearly identify the origin of a commit.

Agreed, this would help in maintaining and doing regular rebases to upstream. But we need to pull in few tools/repos from NCS to make it happen, but I guess we could start with just the `sauce tags`.

jukkar:
We (me, Carles and Robert) pondered the commit tagging internally and found out that:
* Rebasing and rewriting history is not allowed according to this zephyr policy https://docs.zephyrproject.org/latest/develop/modules.html#allowed-practices
* It seems that only `[toup]` and `[noup]` make sense with upstream hostap (those tags could be named also `[zephyr toup]` or `[zephyr noup]`)
* If the commit is accepted to upstream (the `[toup]` tag was used), then when merging the upstream back, we would need to revert the original commit (because the commit message is not the same)

We can certainly do this, if needed. WDYT?

jukkar:
If we want to get the hostap updates here and in the glue PR in zephyr repo merged to 3.6, then we would need to speed up the review as merge window closes next week. @dleach02 please prioritize this.

src/utils/os_zephyr.c
```diff
@@ -191,7 +191,7 @@ int os_fdatasync(FILE *stream)
 
 char *os_strdup(const char *s)
 {
-	return strdup(s);
+	return os_memdup(s, strlen(s) + 1);
```
krish2718, line 194: interesting, may be leave a TODO here?

krish2718 (approved):

dleach02:
> We (me, Carles and Robert) pondered the commit tagging internally and found out that:
> 
> * Rebasing and rewriting history is not allowed according to this zephyr policy [docs.zephyrproject.org/latest/develop/modules.html#allowed-practices](https://docs.zephyrproject.org/latest/develop/modules.html#allowed-practices)
> * It seems that only `[toup]` and `[noup]` make sense with upstream hostap (those tags could be named also `[zephyr toup]` or `[zephyr noup]`)
> * If the commit is accepted to upstream (the `[toup]` tag was used), then when merging the upstream back, we would need to revert the original commit (because the commit message is not the same)
> 
> We can certainly do this, if needed. WDYT?

I'm fine with that. I actually think that is how Nordic typically does this anyway when doing the mergeup; the PR has a bunch of reverts then the pull in of the upstream, followed by reapply of those items that are still pending. But you guys know this process better then I do ;-)

I just feel that this project appears to be going down a path of having lots of changes to the upstream and I think the Nordic workflow will help avoid technical debt in the long run.

dleach02 (approved):

krish2718:
> The base branch requires all commits to be signed.

I guess this was accidental and should be disabled? I don't think Zephyr enforces `signed commits`.

PR #6, opened 2024-04-23
Hostap update 20240423

jukkar:
One commit cherry-picked from Nordic downstream.
Also noticed that `inet_ntoa()` comes now automatically when `POSIX_API` is set so no need to have it separately.


jukkar:
ping @krish2718

krish2718 (approved):

PR #1, opened 2023-10-20
Changed needed for Zephyr

jukkar:
Initial changes needed for compiling with Zephyr.
Corresponding glue code in https://github.com/zephyrproject-rtos/zephyr/pull/64168 will be sent next.

This does not yet contain driver integration code as I will sent it after this one is merged.

PR #7, opened 2024-05-22
zephyr hostap bug fix

MaochenWang1:
Fix EAP packet dropped unexpectedly,
Use zephyr system heap for hostap,
Fix ctrl msg buffer overflow

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -1329,7 +1329,6 @@ static int wpa_drv_zep_associate(void *priv,
 		goto out;
 	}
 
-	ret = dev_ops->associate(if_ctx->dev_priv, params);
 	if (ret) {
 		wpa_printf(MSG_ERROR, "%s: associate op failed", __func__);
 		goto out;
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -1349,7 +1349,8 @@ static int _wpa_drv_zep_set_key(void *priv,
 				const u8 *seq,
 				size_t seq_len,
 				const u8 *key,
-				size_t key_len)
+				size_t key_len,
+				enum key_flag key_flag)
 {
 	struct zep_drv_if_ctx *if_ctx = NULL;
 	const struct zep_wpa_supp_dev_ops *dev_ops;
@@ -1381,7 +1382,7 @@ static int _wpa_drv_zep_set_key(void *priv,
 	}
 
 	wpa_printf(MSG_DEBUG, "%s: priv:%p alg %d addr %p key_idx %d set_tx %d seq %p "
-		   "seq_len %d key %p key_len %d",
+		   "seq_len %d key %p key_len %d key_flag %x",
 		   __func__,
 		   if_ctx->dev_priv,
 		   alg, addr,
@@ -1390,7 +1391,8 @@ static int _wpa_drv_zep_set_key(void *priv,
 		   seq,
 		   seq_len,
 		   key,
-		   key_len);
+		   key_len,
+		   key_flag);
 
 	ret = dev_ops->set_key(if_ctx->dev_priv,
 			       ifname,
@@ -1401,7 +1403,8 @@ static int _wpa_drv_zep_set_key(void *priv,
 			       seq,
 			       seq_len,
 			       key,
-			       key_len);
+			       key_len,
+			       key_flag);
 	if (ret) {
 		wpa_printf(MSG_ERROR, "%s: set_key op failed", __func__);
 		goto out;
@@ -1423,7 +1426,8 @@ static int wpa_drv_zep_set_key(void* priv,
 				    params->seq,
 				    params->seq_len,
 				    params->key,
-				    params->key_len);
+				    params->key_len,
+				    params->key_flag);
 }
 
 
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -914,17 +914,27 @@ wpa_driver_wpa_supp_postprocess_modes(struct hostapd_hw_modes *modes,
 	return modes;
 }
 
-struct hostapd_hw_modes *wpa_drv_get_hw_feature_data(void *priv,
+struct hostapd_hw_modes *wpa_drv_zep_get_hw_feature_data(void *priv,
 		u16 *num_modes,
 		u16 *flags, u8 *dfs_domain)
 {
 	struct zep_drv_if_ctx *if_ctx = NULL;
 	const struct zep_wpa_supp_dev_ops *dev_ops;
+	struct hostapd_hw_modes *modes = NULL;
 	int ret = -1;
 
 	if_ctx = priv;
 
 	dev_ops = get_dev_ops(if_ctx->dev_ctx);
+	if (!dev_ops) {
+		wpa_printf(MSG_ERROR, "%s:Failed to get dev_ops handle", __func__);
+		goto out;
+	}
+
+	if (!dev_ops->get_wiphy) {
+		wpa_printf(MSG_ERROR, "%s: No op registered for get_wiphy", __func__);
+		goto out;
+	}
 
 	struct phy_info_arg result = {
 		.num_modes = num_modes,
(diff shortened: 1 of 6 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -366,6 +366,12 @@ void wpa_drv_zep_event_proc_scan_done(struct zep_drv_if_ctx *if_ctx,
 			event);
 }
 
+void wpa_drv_zep_event_proc_scan_abort(struct zep_drv_if_ctx *if_ctx)
```
krish2718, line 369: This can simply be SCAN_DONE, WPA supplicant doesn't treat this any differently.

src/utils/os_zephyr.c
```diff
@@ -204,9 +215,54 @@ void *os_memdup(const void *src, size_t len)
 	return r;
 }
 
+void *os_malloc(size_t size)
+{
+    return k_malloc(size);
```
krish2718, line 220: So, far WPA supplicant has been treated as an application, so, it used `malloc`, I do agree that for Zephyr this is more of a kernel service, need to think a bit more on this.

krish2718 (commented):
Please split the commits to limit to a single logical change.

src/drivers/driver_zephyr.c
```diff
@@ -366,6 +366,12 @@ void wpa_drv_zep_event_proc_scan_done(struct zep_drv_if_ctx *if_ctx,
...
+{
+    wpa_printf(MSG_DEBUG, "%s: Scan aborted by driver", __func__);
+
+    if_ctx->scan_res2_get_in_prog = false;
```
jukkar, line 373: Please follow zephyr coding style in these files. So use tabs (size 8) instead of spaces.
https://docs.zephyrproject.org/latest/contribute/guidelines.html#coding-style

src/l2_packet/l2_packet_zephyr.c
```diff
@@ -94,9 +94,7 @@ static void l2_packet_receive(int sock, void *eloop_ctx, void *sock_ctx)
 	// FIXME: sll_addr is not being filled as L2 header is not removed
 	hdr = (const struct ieee802_1x_hdr *) buf;
 
-	// FIXME: L2 header is now removed but sll_protocol is not set.
-	// So, as a workaround add this check to drop packets.
-	if (hdr->type != IEEE802_1X_TYPE_EAPOL_KEY)
+	if (ll.sll_protocol != ETH_P_EAPOL)
```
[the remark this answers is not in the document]
dleach02: @jukkar, will the Zephyr Project consider upstreaming changes we make to the hostap back to the original project? If so, do they have a coding style that should be followed?

src/utils/os_zephyr.c
dleach02: since this file is an interface layer between hostap and Zephyr it seems like we should use the k_malloc and other Zephyr kernel primitives.

MaochenWang1: Hi @krish2718 , when test the enterprise mode, for example the eap-tls, it will dynamically allocate large memory, and use malloc is not enough and will lead to connection fail. As the zephyr kernel heap is commonly use and have large size, it's more suitable to use.

src/drivers/driver_zephyr.c
MaochenWang1: thanks, will update

src/l2_packet/l2_packet_zephyr.c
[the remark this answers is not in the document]
MaochenWang1: will update

src/utils/os_zephyr.c
```diff
@@ -204,9 +215,54 @@ void *os_memdup(const void *src, size_t len)
...
+    if (newsize == 0)
+    {
+        os_free(ptr);
+        return NULL;
+    }
+
+    if (ptr == NULL)
+    {
+        return os_malloc(newsize);
+    }
```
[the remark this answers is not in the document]
MaochenWang1: will update

src/drivers/driver_zephyr.c
MaochenWang1: Hi @krish2718  this is the callback function from driver side and it's different case from scan_done.

MaochenWang1:
Hi @jukkar @krish2718 I have updated, pls help review again, thanks

src/l2_packet/l2_packet_zephyr.c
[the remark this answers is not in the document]
jukkar: @dleach02 dunno, but as this is zephyr specific file, we could try to follow our own style at last for time being. We can of course change this later if needed.

src/drivers/driver_zephyr.c
```diff
@@ -1336,7 +1336,6 @@ static int wpa_drv_zep_associate(void *priv,
 		goto out;
```
jukkar, line 1336: Typo in commit subject, `Rremove` should be `Remove`

Also it would be nice to see description why this change is needed in the commit message.

src/drivers/driver_zephyr.c
```diff
@@ -1356,7 +1356,8 @@ static int _wpa_drv_zep_set_key(void *priv,
 				const u8 *seq,
```
jukkar, line 1356: Also in commit `Add key_flag parameter for set_key`, it is missing commit body that should describe why this change is done/needed. As a general note, the commit body should not be left empty.

src/drivers/driver_zephyr.c
MaochenWang1: updated, pls review again

src/drivers/driver_zephyr.c
MaochenWang1: updated, pls review again

jukkar (approved):
Thanks LGTM.

dleach02 (approved):

src/drivers/driver_zephyr.c
krish2718: ok, but how does this plugin to the WPA supplicant? apart from acting as an intimation of SCAN_DONE?

src/utils/os_zephyr.c
krish2718: >  and use malloc is not enough and will lead to connection fail

Well, we could increase the heap.

> As the zephyr kernel heap is commonly use and have large size, it's more suitable to use.

intree Zephyr doesn't have much heap usage, but as David says above, if we treat supplicant as core service of Zephyr, then it's fine to move to `k_heap`.

src/l2_packet/l2_packet_zephyr.c
```diff
@@ -94,10 +94,9 @@ static void l2_packet_receive(int sock, void *eloop_ctx, void *sock_ctx)
 	// FIXME: sll_addr is not being filled as L2 header is not removed
 	hdr = (const struct ieee802_1x_hdr *) buf;
 
-	// FIXME: L2 header is now removed but sll_protocol is not set.
-	// So, as a workaround add this check to drop packets.
```
krish2718, line 98: AFAIK, this is still applicable, no?

krish2718 (changes requested):
All the patches in this PR are Zephyr specific, so, can just be `noup`, we haven't gotten much traction for the initial Zephyr support patch, so, all Zephyr specific changes are `noup`.

src/l2_packet/l2_packet_zephyr.c
MaochenWang1: will add it back

src/drivers/driver_zephyr.c
MaochenWang1: Hi @krish2718 I have updated the wpa_drv_zep_event_proc_scan_abort() function.
This is triggered by driver, as the driver's scan is in progress and aborted (maybe by "wifi disconnect" cmd), driver should notify supplicant, and supplicant can release the scan_work, don't let the next connection is blocked by the scan_work. 
scan done is all the channel scanned over, but scan abort is scan aborted in progress

MaochenWang1:
Hi @krish2718 
have updated as your comment.
about the manifest, will update it after hostp repo PR merged.
Pls review again.

src/drivers/driver_zephyr.c
krish2718: Sure, I understand, as you say it's main purpose is to make the scan free in WPA supplicant which can be done in two ways:

1. Send SCAN_DONE even for driver initiated abort
2. WPA supplicant itself aborts the scan, e.g., disconnect

For #1 I don't think we need to distinguish if it's regular SCAN_DONE or SCAN_DONE from driver abort.

MaochenWang1: okay, I will check and remove this part

MaochenWang1:
Hi @krish2718 
Pls review again, thanks

src/l2_packet/l2_packet_zephyr.c
krish2718: Sorry, didn't check the fix, looks like now L2 sll_protocol is set? If yes, you can remove the FIXME

krish2718 (approved):

src/l2_packet/l2_packet_zephyr.c
MaochenWang1: yes, updated, thanks

MaochenWang1:
Hi @krish2718 @jukkar can you help merge this PR?

gannicus-gj (approved):

gannicus-gj:
@jukkar , @krish2718 , can you merge this PR or guide us anything pending to do?

PR #8, opened 2024-05-24
[toup] zephyr: l2_packet: socket protocol use little endian

fengming-ye:
Zephyr socket protocol is using little endian match. So change l2 packet socket protocol to little endian.

### DIFF
--- a/src/l2_packet/l2_packet_zephyr.c
+++ b/src/l2_packet/l2_packet_zephyr.c
@@ -144,8 +144,8 @@ l2_packet_init(const char *ifname, const u8 *own_addr, unsigned short protocol,
 	link_addr = &iface->if_dev->link_addr;
 	os_memcpy(l2->own_addr, link_addr->addr, link_addr->len);
 
-	l2->fd = socket(AF_PACKET, l2_hdr ? SOCK_RAW : SOCK_DGRAM,
-			htons(protocol));
+	/* zephyr socket protocol is implemented in little-endian */
+	l2->fd = socket(AF_PACKET, l2_hdr ? SOCK_RAW : SOCK_DGRAM, protocol);
 	if (l2->fd < 0) {
 		wpa_printf(MSG_ERROR,
 			   "Failed to open socket: %s, proto: %d, af: %d",


src/l2_packet/l2_packet_zephyr.c
```diff
@@ -144,8 +144,8 @@ l2_packet_init(const char *ifname, const u8 *own_addr, unsigned short protocol,
 	link_addr = &iface->if_dev->link_addr;
 	os_memcpy(l2->own_addr, link_addr->addr, link_addr->len);
 
-	l2->fd = socket(AF_PACKET, l2_hdr ? SOCK_RAW : SOCK_DGRAM,
-			htons(protocol));
+	/* zephyr socket protocol is implemented in little-endian */
```
jukkar, line 147: This is not exactly correct. AFAICT, the protocol field should be host byte order when calling the `socket()`, be it either big or little endian.
I just wonder how this has been working at all earlier, unless the protocol field is already in network byte order and the code just swapped it back to host byte order which is expected by the `socket()` call. In which case this change would not work, how well have you tested the code after this change?
@krish2718 any idea what is going on here?

jukkar: I checked the Linux specific code, and it is doing things exactly like this, so it kinds of implies that the protocol value is already in network byte order.

jukkar: Few lines below there is a line `ll.sll_protocol = htons(protocol);` and that tells that protocol is indeed in host byte order.
Found this on this manual page https://www.man7.org/linux/man-pages/man7/packet.7.html
```
When protocol is set to htons(ETH_P_ALL), then all protocols are received.
```
So this tells that the this fix is not correct but we need to fix this in packet socket code.

fengming-ye: zephyr other packet sockets are all using host order protocol
so either change all zephyr packet sockets create code or change supplicant code.

fengming-ye: eg. in zephyr/tests/net/socket/af_packet/src/main.c

jukkar: We need to change all zephyr packet socket create code in this case, more work but in order to be interoperable with other OS this needs to be done like this. Also this change needs to be documented to migration guide.

dleach02: @jukkar, Are you suggesting that we need to go change all of the code creating or interfacing with sockets to us htons/ntohs for all protocol fields? 

Who should do this? 

And do we then table this particular PR in preference to the global change to align the socket interaction code to align with using network/host byte ordering macros?

jukkar: This change should be quite straight forward, I can certainly do it if needed if you are busy with other stuff.

fengming-ye: Hi @jukkar  That will be greate if you can do the AF_PACKET socket improvement, as I'm upstreaming crypto_mbedtls and need to ensure it does not break your supplicant work.
The zephyr repo fix supplicant l2 unexpected frame commit can be hold until your PR merged.
https://github.com/zephyrproject-rtos/zephyr/pull/73120

jukkar: Fix proposal for the `ETH_P_ALL` issue can be found in https://github.com/zephyrproject-rtos/zephyr/pull/73338

PR #9, opened 2024-05-24
[toup] zephyr: crypto: port crypto mbedtls wrapper

fengming-ye:
Port crypto mbedtls wrapper for enterprise and DPP.

Porting from
https://github.com/gstrauss/hostap/blob/mbedtls/src/crypto/crypto_mbedtls.c

### DIFF
--- a/port/mbedtls/supp_psa_api.c
+++ b/port/mbedtls/supp_psa_api.c
@@ -1,0 +1,419 @@
+/*
+ *  Copyright The Mbed TLS Contributors
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/*
+ *  Copyright 2023-2024 NXP
+ *
+ *  @file  supp_psa_api.c
+ *  @brief This file provides wpa supplicant crypto mbedtls PSA APIs.
+ */
+
+#include "supp_psa_api.h"
+
+#define ASSERT_STATUS(actual, expected)                                            \
+    do                                                                             \
+    {                                                                              \
+        if ((actual) != (expected))                                                \
+        {                                                                          \
+            printk(                                                                \
+                "\tassertion failed at %s:%d - "                                   \
+                "actual:%d expected:%d\r\n",                                       \
+                __FILE__, __LINE__, (psa_status_t)actual, (psa_status_t)expected); \
+            goto exit;                                                             \
+        }                                                                          \
+    } while (0)
+
+#define SUPP_PSA_MAX_OUTPUT_SIZE 2048
+
+static uint8_t supp_psa_outbuf[SUPP_PSA_MAX_OUTPUT_SIZE];
+
+static inline void supp_psa_set_attributes(psa_key_attributes_t *attributes, u32 type, u32 alg, u32 usage)
+{
+    psa_set_key_type(attributes, type);
+    psa_set_key_algorithm(attributes, alg);
+    psa_set_key_usage_flags(attributes, usage);
+}
+
+static void supp_psa_get_hash_alg(mbedtls_md_type_t type, psa_algorithm_t *alg, int *block_size)
+{
+    switch (type)
+    {
+        case MBEDTLS_MD_MD5:
+            *alg = PSA_ALG_MD5;
+            break;
+        case MBEDTLS_MD_SHA1:
+            *alg = PSA_ALG_SHA_1;
+            break;
+        case MBEDTLS_MD_SHA224:
+            *alg = PSA_ALG_SHA_224;
+            break;
+        case MBEDTLS_MD_SHA256:
+            *alg = PSA_ALG_SHA_256;
+            break;
+        case MBEDTLS_MD_SHA384:
+            *alg = PSA_ALG_SHA_384;
+            break;
+        case MBEDTLS_MD_SHA512:
+            *alg = PSA_ALG_SHA_512;
+            break;
+        case MBEDTLS_MD_RIPEMD160:
+            *alg = PSA_ALG_RIPEMD160;
+            break;
+        default:
+            *alg = PSA_ALG_NONE;
+            break;
+    }
+    *block_size = PSA_HASH_LENGTH(*alg);
+}
+
+static psa_status_t supp_psa_cipher_operation(psa_cipher_operation_t *operation,
+                                              const uint8_t *input,
+                                              size_t input_size,
+                                              size_t part_size,
+                                              uint8_t *output,
+                                              size_t output_size,
+                                              size_t *output_len)
+{
+    psa_status_t status;
+    size_t bytes_to_write = 0;
+    size_t bytes_written  = 0;
+    size_t len            = 0;
+
+    *output_len = 0;
+    while (bytes_written != input_size)
+    {
+        bytes_to_write = (input_size - bytes_written > part_size ? part_size : input_size - bytes_written);
+
+        status = psa_cipher_update(operation, input + bytes_written, bytes_to_write, output + *output_len,
+                                   output_size - *output_len, &len);
+        ASSERT_STATUS(status, PSA_SUCCESS);
+
+        bytes_written += bytes_to_write;
+        *output_len += len;
+    }
+
+    status = psa_cipher_finish(operation, output + *output_len, output_size - *output_len, &len);
+    ASSERT_STATUS(status, PSA_SUCCESS);
+    *output_len += len;
+
+exit:
+    return status;
+}
+
+#ifdef MBEDTLS_AES_C
+#define SUPP_PSA_AES_BLOCK_SIZE 16
+
+int aes_128_encrypt_block_psa(const u8 *key, const u8 *in, u8 *out)
+{
+    psa_status_t status;
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    mbedtls_svc_key_id_t key_id     = MBEDTLS_SVC_KEY_ID_INIT;
+    psa_algorithm_t alg             = PSA_ALG_ECB_NO_PADDING;
+    size_t out_len                  = 0;
+
+    supp_psa_set_attributes(&attributes, PSA_KEY_TYPE_AES, alg, PSA_KEY_USAGE_ENCRYPT);
+
+    status = psa_import_key(&attributes, key, SUPP_PSA_BLOCK_SIZE_128, &key_id);
+    ASSERT_STATUS(status, PSA_SUCCESS);
+    psa_reset_key_attributes(&attributes);
+
+    status = psa_cipher_encrypt(key_id, alg, in, SUPP_PSA_BLOCK_SIZE_128, out, SUPP_PSA_BLOCK_SIZE_128, &out_len);
+    ASSERT_STATUS(status, PSA_SUCCESS);
+
+exit:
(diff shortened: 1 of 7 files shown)


src/crypto/crypto_mbedtls.c
```diff
@@ -1,911 +1,4042 @@
 /*
- * SPDX-FileCopyrightText: 2020-2021 Espressif Systems (Shanghai) CO LTD
- * SPDX-FileCopyrightText: 2022 Nordic Semiconductor ASA
+ * crypto wrapper functions for mbed TLS
  *
- * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: 2022 Glenn Strauss <gstrauss@gluelogic.com>
+ * SPDX-License-Identifier: BSD-3-Clause
```
jukkar, line 5: You just cannot change the license and remove the old copyright statements, this is not how open source works.
cc: @dleach02

jukkar (changes requested):

src/crypto/crypto_mbedtls.c
fengming-ye: it is total re-write. could you pls specify more how I shall do this, if current PR is incorrect?

fengming-ye:
Hi @jukkar @krish2718 
crypto_mbedtls.c and tls_mbedtls.c are both total re-write and replace.
is this okay to merge?

src/crypto/crypto_mbedtls.c
krish2718: AFAIK, `Apache-2.0` doesn't allow to relicense the original code, the new code can be under new licesne but don't quote me on that, need some legal help here.

Do you want your changes to be `BSD-3-clause` only? If yes, then another way would be to a move to `port/mbedtls/crypto_mbedtls.c` and add a Kconfig to choose different implementations?

BTW, what about remaining files `crypto_mbedtls-ec.c`, `crypto_mbedtls-bignum.c` ...? Does your implementation don't need them?

krish2718: > I can add new BSD license under origin license. Is that looks okay to you?

Not sure if that would be allowed or not, others can chime in.

> crypto_mbedtls-ec.c, crypto_mbedtls-bignum.c I don't need them.

But WPA supplicant calls `crypto_bignum_*` APIs esp. in `sae.c` and `sae_pk.c` and also `crypto_ec_*` from `dpp.c` etc. How does this work?

fengming-ye: they are in crypto_mbedtls.c

fengming-ye: we are using like crypto_linux.c, maintaining them in one file.
and about the license, this PR is misleading. Because I removed it and replaced by a new file, just with the same name, but no same content.
Shall I raise a commit to delete it first and then create a new one, in split commits?
Or shall I name the new one crypto_mbedtls2.c, and keep origin file?

krish2718: Ah ok, missed that. So, I would definitely recommend submitting this as an alternative implementation:

```
backend_a/CURRENT MbedTLS
backend_b/This MBEDTLS

Kconfig: 

choice MBEDTLS_BACKEND
   
config a
config b
```

Not quite sure about naming of `a/b`, this way you are free to use your own license? @jukkar Is there a policy in Zephyr that `Apache-2.0` is the preferred or mandated License?

fengming-ye: > Ah ok, missed that. So, I would definitely recommend submitting this as an alternative implementation:
> 
> ```
> backend_a/CURRENT MbedTLS
> backend_b/This MBEDTLS
> 
> Kconfig: 
> 
> choice MBEDTLS_BACKEND
>    
> config a
> config b
> ```
> 

I agree with this. Is there a recommend name?

krish2718: Well, that is a tough decision, I don't want to use company names like `esp` and `nxp`, but then I don't know what to call them. May be `apache` or `bsd3`? but even that is undesirable. Let me think, meanwhile let's see if someone has any bright ideas here :).

fengming-ye: crypto_mbedtls2.c
crypto_mbedtls_alt.c

does either one look okay?

krish2718: Rather than renaming files, let's to folders, easier to fix in `cmake`.

We can start with `default` and `alt` if no one has any other suggestions.

fengming-ye: you mean create two folders: mbedtls_default and mbedtls_alt. and naming both file crypto_mbedtls.c?
I think I can add file and kconfig like

zephyr_library_sources_ifdef(CONFIG_WIFI_NM_WPA_SUPPLICANT_CRYPTO_DEFAULT
    ${HOSTAP_SRC_BASE}/crypto/crypto_mbedtls.c
    ${HOSTAP_SRC_BASE}/crypto/tls_mbedtls.c
)

zephyr_library_sources_ifdef(CONFIG_WIFI_NM_WPA_SUPPLICANT_CRYPTO_ALT
    ${HOSTAP_SRC_BASE}/crypto/crypto_mbedtls_alt.c
    ${HOSTAP_SRC_BASE}/crypto/tls_mbedtls_alt.c
)

This looks easier and okay?

krish2718: Yes and for `CRYPTO_DEFAULT` we cann bignum, ec etc files.

fengming-ye:
Update as previous comments.
Add crypto_mbedtls_alt.c and tls_mbedtls_alt.c enabled by CONFIG_WIFI_NM_WPA_SUPPLICANT_CRYPTO_ALT
in this PR
https://github.com/zephyrproject-rtos/zephyr/pull/73249

dleach02:
Are we good with the copyright concerns now?

jukkar:
> Are we good with the copyright concerns now?

Yes, looks much better now, thanks for the fixes.

jukkar (approved):

dleach02 (approved):

jukkar:
@krish2718 please take a look

krish2718:
LGTM, except for last commit, can you please elaborate on `Thus when DPP enabled, unfixed warning will pop out`? The fix for warnings are essential to pass tests with `-Werror` and the data type fix was found using `coverity` (downstream NCS).

fengming-ye:
> LGTM, except for last commit, can you please elaborate on `Thus when DPP enabled, unfixed warning will pop out`? The fix for warnings are essential to pass tests with `-Werror` and the data type fix was found using `coverity` (downstream NCS).

Sry I didn't catch your point. The last commit is to fix warnings below:

modules/lib/hostap/src/common/dpp.c:3293:37: warning: unsigned conversion from 'int' to 'enum dpp_status_error' changes value from '256' to '0' [-Woverflow]
modules/lib/hostap/src/common/wpa_ctrl.h:208:24: warning: format '%x' expects argument of type 'unsigned int', but argument 4 has type 'long unsigned int' [-Wformat=]

krish2718:
> > LGTM, except for last commit, can you please elaborate on `Thus when DPP enabled, unfixed warning will pop out`? The fix for warnings are essential to pass tests with `-Werror` and the data type fix was found using `coverity` (downstream NCS).
> 
> Sry I didn't catch your point. The last commit is to fix warnings below:
> 
> modules/lib/hostap/src/common/dpp.c:3293:37: warning: unsigned conversion from 'int' to 'enum dpp_status_error' changes value from '256' to '0' [-Woverflow] modules/lib/hostap/src/common/wpa_ctrl.h:208:24: warning: format '%x' expects argument of type 'unsigned int', but argument 4 has type 'long unsigned int' [-Wformat=]

My bad not the last commit, it's this https://github.com/zephyrproject-rtos/hostap/pull/9/commits/00c11d0071655a18e575b71590b6dacfc8079b70 revert.

fengming-ye:
> > > LGTM, except for last commit, can you please elaborate on `Thus when DPP enabled, unfixed warning will pop out`? The fix for warnings are essential to pass tests with `-Werror` and the data type fix was found using `coverity` (downstream NCS).
> > 
> > 
> > Sry I didn't catch your point. The last commit is to fix warnings below:
> > modules/lib/hostap/src/common/dpp.c:3293:37: warning: unsigned conversion from 'int' to 'enum dpp_status_error' changes value from '256' to '0' [-Woverflow] modules/lib/hostap/src/common/wpa_ctrl.h:208:24: warning: format '%x' expects argument of type 'unsigned int', but argument 4 has type 'long unsigned int' [-Wformat=]
> 
> My bad not the last commit, it's this [00c11d0](https://github.com/zephyrproject-rtos/hostap/commit/00c11d0071655a18e575b71590b6dacfc8079b70) revert.

Oh this is because except from some printfs %ld changed to %lld. There are more %ld unchanged. And I don't want to change much DPP original code for this. I have doubt that if %ld is changed to %lld, won't it cause type warning on Linux?

On the other hand, why would os_time_t defined in unsigned long has warnings as Linux is using unsigned long?
Could you pls share the warning?

BTW my local build has no warning with these commits.

krish2718:
> > > > LGTM, except for last commit, can you please elaborate on `Thus when DPP enabled, unfixed warning will pop out`? The fix for warnings are essential to pass tests with `-Werror` and the data type fix was found using `coverity` (downstream NCS).
> > > 
> > > 
> > > Sry I didn't catch your point. The last commit is to fix warnings below:
> > > modules/lib/hostap/src/common/dpp.c:3293:37: warning: unsigned conversion from 'int' to 'enum dpp_status_error' changes value from '256' to '0' [-Woverflow] modules/lib/hostap/src/common/wpa_ctrl.h:208:24: warning: format '%x' expects argument of type 'unsigned int', but argument 4 has type 'long unsigned int' [-Wformat=]
> > 
> > 
> > My bad not the last commit, it's this [00c11d0](https://github.com/zephyrproject-rtos/hostap/commit/00c11d0071655a18e575b71590b6dacfc8079b70) revert.
> 
> Oh this is because except from some printfs %ld changed to %lld. There are more %ld unchanged. And I don't want to change much DPP original code for this. I have doubt that if %ld is changed to %lld, won't it cause type warning on Linux?
> 
The fixes weren't upstreamed but this all depends on the host machine that you are building for.

> On the other hand, why would os_time_t defined in unsigned long has warnings as Linux is using unsigned long? Could you pls share the warning?

This was warning from coverity, and it should be signed 64bit to capture full times resolution.

I need to dig my mailbox for the warnings, but I would suggest to remove this commit from the PR and we can handle it separately as it needs a discussion. Rest of the commits are good to go. WDYT?
> 
> BTW my local build has no warning with these commits.

Yeah, might be toolchain dependent and which flags are enabled by default.

fengming-ye:
> > > > > LGTM, except for last commit, can you please elaborate on `Thus when DPP enabled, unfixed warning will pop out`? The fix for warnings are essential to pass tests with `-Werror` and the data type fix was found using `coverity` (downstream NCS).
> > > > 
> > > > 
> > > > Sry I didn't catch your point. The last commit is to fix warnings below:
> > > > modules/lib/hostap/src/common/dpp.c:3293:37: warning: unsigned conversion from 'int' to 'enum dpp_status_error' changes value from '256' to '0' [-Woverflow] modules/lib/hostap/src/common/wpa_ctrl.h:208:24: warning: format '%x' expects argument of type 'unsigned int', but argument 4 has type 'long unsigned int' [-Wformat=]
> > > 
> > > 
> > > My bad not the last commit, it's this [00c11d0](https://github.com/zephyrproject-rtos/hostap/commit/00c11d0071655a18e575b71590b6dacfc8079b70) revert.
> > 
> > 
> > Oh this is because except from some printfs %ld changed to %lld. There are more %ld unchanged. And I don't want to change much DPP original code for this. I have doubt that if %ld is changed to %lld, won't it cause type warning on Linux?
> 
> The fixes weren't upstreamed but this all depends on the host machine that you are building for.
> 
> > On the other hand, why would os_time_t defined in unsigned long has warnings as Linux is using unsigned long? Could you pls share the warning?
> 
> This was warning from coverity, and it should be signed 64bit to capture full times resolution.
> 
> I need to dig my mailbox for the warnings, but I would suggest to remove this commit from the PR and we can handle it separately as it needs a discussion. Rest of the commits are good to go. WDYT?
> 
> > BTW my local build has no warning with these commits.
> 
> Yeah, might be toolchain dependent and which flags are enabled by default.

Okay I removed this commit.

krish2718 (approved):

PR #10, opened 2024-05-31
Feature/dpp support

fengming-ye:
[toup] dpp: add zephyr own timeout for dpp response timeout
    
    For some MCUs in zephyr that has lite CPU, hard code 1s or 2s may be
    not enough for SHA384 and SHA512 crypto calculation.
    
    So extend DPP wait for reponse timeout to 5s.

[toup] config: add dpp_config_processing
    
    Add dpp_config_processing 2, which means
    report received configuration to an external program, generate
    a network profile internally, try to connect to the created
    profile automatically.

[toup] zephyr: fix wpa_msg always use fixed msg size
    
    wpa_cli_open_connection create a mon_sock_pair to monitor hostap events.
    wpa_msg send msgs always with fixed size, 256 bytes for now.
    And wpa_cli_recv_pending receive msg through k_pipe_get by spair_read,
    which has, eg. 4KB queue size.
    
    That means if hostap eloop send more than 16 msgs, the socket pair will
    be full and block. Thus eloop cannot read socket pair and will be forever
    blocked. This case happens easy in DPP case as it has many wpa_msg INFO.
    
    Fix it by read msg header with variable length first,
    and then read whole msg.
    This will make more use of mon_sock_pair queue size.

### DIFF
--- a/wpa_supplicant/wpa_cli_zephyr.c
+++ b/wpa_supplicant/wpa_cli_zephyr.c
@@ -118,9 +118,19 @@ static void wpa_cli_recv_pending(struct wpa_ctrl *ctrl, struct wpa_supplicant *w
 {
 	while (wpa_ctrl_pending(ctrl) > 0) {
 		char buf[sizeof(struct conn_msg)];
-		size_t len = sizeof(buf);
+		size_t hlen = sizeof(int);
+		size_t plen = MAX_CTRL_MSG_LEN;
+		int need_flush = 0;
 
-		if (wpa_ctrl_recv(ctrl, buf, &len) == 0) {
+		if (wpa_ctrl_recv(ctrl, buf, &hlen) == 0 &&
+		    hlen == sizeof(int)) {
+			plen = *((int *)buf);
+		} else {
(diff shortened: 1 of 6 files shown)


jukkar (approved):
LGTM

dleach02 (approved):

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -118,14 +118,24 @@ static void wpa_cli_recv_pending(struct wpa_ctrl *ctrl, struct wpa_supplicant *w
...
+			wpa_printf(MSG_ERROR, "Could not read pending message header len %d.\n", hlen);
+			continue;
+		}
+
+		if (wpa_ctrl_recv(ctrl, buf + sizeof(int), &plen) == 0) {
 			struct conn_msg *msg = (struct conn_msg *)buf;
 
 			msg->msg[msg->msg_len] = '\0';
 			wpa_printf(MSG_DEBUG, "Received len: %d, msg_len:%d - %s->END\n",
-					   len, msg->msg_len, msg->msg);
```
krish2718, line 128: unnecesary whitespace change

krish2718 (approved):

PR #12, opened 2024-06-07
Feature/wpa cli cmd resp

fengming-ye:
[toup] zephyr: parse wpa_cli cmd resp without LF
    
    Some wpa_cli command responses end with LF, but some not.
    eg. DPP command responses are "1".
    
    So remove only LF when the 2nd last character is LF.
    
    Signed-off-by: Fengming Ye <frank.ye@nxp.com>

[toup] zephyr: add wpa_cli command with response
    
    Add wpa_cli command with response, in case some commands need to
    get response.
    eg. DPP commands print response info like bootstrap uri and added
    configurator id, bootstrap id, etc.
    But we want to parse them in L2 mgmt runtime function call instead of
    on console log.
    
    Signed-off-by: Fengming Ye <frank.ye@nxp.com>

### DIFF
--- a/wpa_supplicant/wpa_cli_zephyr.c
+++ b/wpa_supplicant/wpa_cli_zephyr.c
@@ -90,6 +90,11 @@ static int wpa_ctrl_command_resp(struct wpa_ctrl *ctrl, const char *cmd, char *r
 	return _wpa_ctrl_command(ctrl, cmd, 0, resp);
 }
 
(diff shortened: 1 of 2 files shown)


dleach02 (approved):

krish2718 (commented):
As we don't have Zephyr support yet in upstream, please mark them as `noup`, if at all it's added in the future we can update the tag.

gannicus-gj (approved):

jukkar (approved):

PR #13, opened 2024-06-20
[noup] zephyr: use MSG_INFO level for wpa_cli command response print

fengming-ye:
wpa_cli command with interactive means to show response info on console.
In this case it is better to use MSG_INFO level to print instead of
MSG_DEBUG level. As MSG_DEBUG level is for debug cases, but wpa_cli
response print is normal use case.

### DIFF
--- a/wpa_supplicant/wpa_cli_zephyr.c
+++ b/wpa_supplicant/wpa_cli_zephyr.c
@@ -83,7 +83,7 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print, 
 	if (print) {
 		buf[len] = '\0';
 		if (buf[0] != '\0')
-			wpa_printf(MSG_DEBUG, "%s", buf);
+			wpa_printf(MSG_INFO, "%s", buf);
 	}
 
 	return 0;


jukkar (approved):

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -83,7 +83,7 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print,
 	if (print) {
 		buf[len] = '\0';
 		if (buf[0] != '\0')
-			wpa_printf(MSG_DEBUG, "%s", buf);
+			wpa_printf(MSG_INFO, "%s", buf);
```
krish2718, line 86: This would mean more noise for other commands, instead of print can't we prepare a structure with response and send it back? See `z_wpa_ctrl_status` as an example.

fengming-ye: I don't think response info is noise.
I think command with interactive means they need to get response shown.
DEBUG level means this log is for debug, but I think this log is for user info.

krish2718: In the upstream, wpa_cli is an interactive application that uses control interface, but in Zephyr downstream the control interface is not interactive, so, all responses are parsed and returned to the interested caller via structures programatically.

fengming-ye: Changed the commit message to try to solve the misleading here.

krish2718: Thanks for clarifying.

krish2718 (approved):

krish2718:
can you please raise a corresponding manifest update PR?

fengming-ye:
> can you please raise a corresponding manifest update PR?

Sure raised
https://github.com/zephyrproject-rtos/zephyr/pull/74599

PR #16, opened 2024-07-18
[noup] zephyr: Add chan_list_changed API

LiLongNXP:
Add chan_list_changed API

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -289,6 +289,13 @@ void wpa_supplicant_event_wrapper(void *ctx,
 	zephyr_wifi_send_event(&msg);
 }
 
+void wpa_drv_zep_event_chan_list_changed(struct zep_drv_if_ctx *if_ctx, union wpa_event_data *event)
+{
+        wpa_supplicant_event_wrapper(if_ctx->supp_if_ctx,
+                EVENT_CHANNEL_LIST_CHANGED,
+                event);
+}
+
 void wpa_drv_zep_event_mac_changed(struct zep_drv_if_ctx *if_ctx)
 {
 	wpa_supplicant_event_wrapper(if_ctx->supp_if_ctx,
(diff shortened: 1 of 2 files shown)


jukkar (approved):
LGTM

gannicus-gj (approved):

gannicus-gj:
Hi @jukkar , @krish2718 
Cince both of you approved this PR, can this PR be merged?

krish2718:
@gannicus-gj there is no manifest PR for this, please raise one. And currently, merging is done in batches as there are too many PR right after the release.

jukkar:
@krish2718 please take a look this PR

krish2718 (approved):

src/drivers/driver_zephyr.c
```diff
@@ -289,6 +289,13 @@ void wpa_supplicant_event_wrapper(void *ctx,
 	zephyr_wifi_send_event(&msg);
 }
 
+void wpa_drv_zep_event_chan_list_changed(struct zep_drv_if_ctx *if_ctx, union wpa_event_data *event)
+{
+        wpa_supplicant_event_wrapper(if_ctx->supp_if_ctx,
+                EVENT_CHANNEL_LIST_CHANGED,
+                event);
```
krish2718, line 296: this needs to be deep-copied in `wpa_supplicant_event_wrapper` else if the driver frees this up and then WPA supplicant accesses it, it will crash.

krish2718 (changes requested):

src/drivers/driver_zephyr.c
LiLongNXP: hi @krish2718 ,
Thanks for your reminder.
I checked the code, the input parameter event's buffer is on stack memory from driver, the fields in this struct memory are all under this stack memory, there is no pointer point to any extra IE in it. 
After event parameter delivered into wpa_supplicant_event_wrapper(), the event is copied to supplicant
		msg.data = os_zalloc(sizeof(*data));
		if (!msg.data) {
			wpa_printf(MSG_ERROR, "Failed to allocated for event: %d", event);
			return;
		}
		os_memcpy(msg.data, data, sizeof(*data));
Then I think this change is ok.
Thanks.
Li Long

krish2718: Event is copied, but it's a shallow-copy, so, any pointers e.g., `ies` won't be copied.

LiLongNXP: hi @krish2718 , 
in union wpa_event_data, this event only fill below fields:
	/**
	 * channel_list_changed - Data for EVENT_CHANNEL_LIST_CHANGED
	 * @initiator: Initiator of the regulatory change
	 * @type: Regulatory change type
	 * @alpha2: Country code (or "" if not available)
	 */
	struct channel_list_changed {
		enum reg_change_initiator initiator;
		enum reg_type type;
		char alpha2[3];
	} channel_list_changed;

There is no pointer in it for extra IEs, so I think no need extra copy for this event.

krish2718 (approved):

PR #19, opened 2024-07-25
[noup] zephyr: set a higher priority for l2_packet

gangli02:
Issue: while running UDP-TX traffic, STAUT receives AP group rekey M1 but cannot send M2, resulting in AP deauth STAUT.
Set the protocol-defined priority of l2_packet sent to 6 (NET_PRIORITY_IC, Internetwork control).
The high priority (>= NET_PRIORITY_CA, 3) packet will be pushed directly to network driver and will skip the traffic class queues.
This will only be true if NET_CONTEXT_PRIORITY and NET_TC_SKIP_FOR_HIGH_PRIO are enabled.

### DIFF
--- a/src/l2_packet/l2_packet_zephyr.c
+++ b/src/l2_packet/l2_packet_zephyr.c
@@ -166,6 +166,22 @@ l2_packet_init(const char *ifname, const u8 *own_addr, unsigned short protocol,
 		goto fail;
 	}
 
+	/* Set the protocol-defined priority of all packets sent on this socket to 6
+	 * (NET_PRIORITY_IC, Internetwork control).
+	 * Network packet priority settings described in IEEE 802.1Q Annex I.1
+	 * The high priority (>= NET_PRIORITY_CA, 3) packet will be pushed directly to network driver
+	 * and will skip the traffic class queues.(Setting a priority outside the range 3 to 6.)
+	 * NOTE: This will only be true if NET_CONTEXT_PRIORITY and NET_TC_SKIP_FOR_HIGH_PRIO are enabled.
+	 */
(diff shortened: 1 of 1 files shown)


src/l2_packet/l2_packet_zephyr.c
```diff
@@ -166,6 +166,21 @@ l2_packet_init(const char *ifname, const u8 *own_addr, unsigned short protocol,
 		goto fail;
 	}
 
+	/* Set the protocol-defined priority of all packets sent on this socket to 6
+	 * (NET_PRIORITY_IC, Internetwork control).
+	 * Network packet priority settings described in IEEE 802.1Q Annex I.1
+	 * The high priority (>= NET_PRIORITY_CA, 3) packet will be pushed directly to network driver
```
krish2718, line 172: This will only be true if `NET_TC_SKIP_FOR_HIGH_PRIO` is enabled, so, please adjust the comment or enable this by default. But none the less, setting high priority is a good idea esp. for rekeying.

krish2718 (approved):
Please

src/l2_packet/l2_packet_zephyr.c
gangli02: Thanks! 
I have adjusted the comments and added a check if NET_TC_SKIP_FOR_HIGH_PRIO is enabled.

gannicus-gj (approved):

src/l2_packet/l2_packet_zephyr.c
krish2718: I was only recommending to modify the comments, independent of that option we can still set a high priority.

gangli02: OK, I see.

PR #24, opened 2024-07-30
[noup] rrm: support custom RRM report length limit

MaochenWang1:
For MBO beacon report test case, if there are too many matched neighbor AP, the RRM measurement report frame length may exceed 2K bytes by default, and this packet will be dropped in wifi driver as it's too long and exceeds the drive buffer size. Now add this custom length config to avoid the packet got dropped

### DIFF
--- a/wpa_supplicant/rrm.c
+++ b/wpa_supplicant/rrm.c
@@ -451,7 +451,11 @@ static void wpas_rrm_send_msr_report(struct wpa_supplicant *wpa_s,
 	int len = wpabuf_len(buf);
 	u8 *pos = wpabuf_mhead_u8(buf), *next = pos;
 
+#ifdef CONFIG_RRM_MSR_REPORT_LEN
+#define MPDU_REPORT_LEN (int) (CONFIG_RRM_MSR_REPORT_LEN - IEEE80211_HDRLEN - 3)
+#else
 #define MPDU_REPORT_LEN (int) (IEEE80211_MAX_MMPDU_SIZE - IEEE80211_HDRLEN - 3)
+#endif
 
 	while (len) {
 		int send_len = (len > MPDU_REPORT_LEN) ? next - pos : len;


MaochenWang1:
Hi @jukkar @krish2718  please review, thanks

wpa_supplicant/rrm.c
```diff
@@ -451,7 +451,11 @@ static void wpas_rrm_send_msr_report(struct wpa_supplicant *wpa_s,
 	int len = wpabuf_len(buf);
 	u8 *pos = wpabuf_mhead_u8(buf), *next = pos;
 
+#ifdef CONFIG_RRM_MSR_REPORT_LEN
+#define MPDU_REPORT_LEN (int) (CONFIG_RRM_MSR_REPORT_LEN - IEEE80211_HDRLEN - 3)
```
krish2718, line 455: This is purely a driver limitation, so, we need to handle it in the driver layer? Using compile time flag isn't right for this, e.g., two drivers using WPA supplicant with different limitations? Can't this be handled in `driver_zephyr.c` in `send_action`  based on driver capability ?

MaochenWang1: Hi @krish2718 I don't think it can be handled in driver_zephyr.c, as the packet is already made and has a length over 2K in wpas_rrm_send_msr_report, it's difficult to split packet into two small one in wpa_drv_zep_send_action().
so limit the length in the source is an easy way to do.

krish2718: Ideally, `IEEE80211_MAX_MMPDU_SIZE` driver should be able to handle this size `2034` as it's the Wi-Fi MTU, if a specific driver is unable to do that, then this should be handled at driver level than a workaround in WPA supplicant. Can't you split the packet in to multiple buffers at driver level itself but still treat it as a single packet?

MaochenWang1: Hi @krish2718  may I ask, the nordic wifi driver and fw supports send the management frame with length 2304?
as #define IEEE80211_MAX_MMPDU_SIZE 2304.

MaochenWang1: Hi @krish2718 if nordic also can't handle the management frame with length 2304, maybe adding CONFIG_RRM_MSR_REPORT_LEN is a common handle?

krish2718: We have the MTU as a configurable option in Nordic driver and by default its 1500, but you can set it to 2304 and then Action frame with 2304 will work.

MaochenWang1: okay, but by default, this action frame with length 2K will be dropped or cause overflow in Nordic driver too, right? and it's difficult to split the action frame in driver level. Adding this macro to distinguish is just a simple way to do.

krish2718: Sure, but it's easily fixable by `CONFIG_NRF_WIFI_IFACE_MTU=2304`, see below the definition (applicable for all Wi-Fi chips), we don't have to add workarounds for driver bugs in the WPA supplicant, esp. 2304 is mandatory as per 802.11.

```
config NRF_WIFI_IFACE_MTU
	int "MTU for Wi-Fi interface"
	range 576 2304 if NET_IPV4
	range 1280 2304 if NET_IPV6
	default 1500
```

MaochenWang1: Got it, thanks for your explain

MaochenWang1:
Hi @jukkar let me think, ,maybe decline this PR in the future

PR #26, opened 2024-08-12
[nrf noup] zephyr: Add support for fetching inactivity timer

krish2718:
In SAP mode, need to fetch the inactivity timer value for a STA from the UMAC.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -2063,6 +2063,29 @@ out:
 	return ret;
 }
 
+
+int wpa_drv_zep_get_inact_sec(void *priv, const u8 *addr)
+{
+	struct zep_drv_if_ctx *if_ctx = priv;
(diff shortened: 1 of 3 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -2063,6 +2063,29 @@ int wpa_drv_hapd_send_eapol(void *priv, const u8 *addr, const u8 *data, size_t d
 	return ret;
```
jukkar, line 2063: The `Fixes SHEL-2681.` seems to be redundant in the commit message as it is only referring to Nordic internal systems.

krish2718: Yes, it was cherry-picked without manual changes,  I can remove that if needed.

krish2718: removed, also removed `nrf` from the sauce-tag.

jukkar (approved):

PR #29, opened 2024-08-20
[toup] zephyr: fix the build error that unknown FILE

MaochenWang1:
When build the Matter over Wi-Fi with NEWLIB enabled, there is build error shows 'error: unknown type name 'FILE' ', adding this change can fix this build error, and don't affect the case that NEWLIB disabled.

### DIFF
--- a/src/utils/os.h
+++ b/src/utils/os.h
@@ -11,6 +11,9 @@
 
 #if defined(__ZEPHYR__)
 typedef long long os_time_t;
+#ifndef FILE
+typedef __FILE FILE;
+#endif
 #else
 typedef long os_time_t;
 #endif


MaochenWang1:
Hi @jukkar @krish2718 @rlubos  please review this change, thanks

jukkar (approved):

krish2718:
> adding this change can
fix this build error, and don't affect the case that NEWLIB disabled.

Shouldn't we do in case some libc defines this?

```
#ifndef FILE
typedef __FILE FILE;
#endif
```

MaochenWang1:
> Shouldn't we do in case some libc defines this?
> 
> ```
> #ifndef FILE
> typedef __FILE FILE;
> #endif
> ```

Hi @krish2718 , thanks, please review again

krish2718 (approved):

dleach02 (approved):

jukkar:
> Shouldn't we do in case some libc defines this?
> 
> ```
> #ifndef FILE
> typedef __FILE FILE;
> #endif
> ```

This only works if `FILE` is defined as a preprocessor symbol but at least for picolibc it is just a typedef for `__FILE`
In picolib we could check `__FILE_defined` and in Linux libc headers it is `____FILE_defined`
Fortunately I could compile this example program just fine without any errors so the cpp check is not really needed (it does no harm either)
```
#include <stdio.h>

void main(void)
{
#ifndef FILE
	printf("FILE is not defined\n");
	typedef __FILE FILE;
#endif

	FILE bar;

	printf("sizeof(FILE) %zd\n", sizeof(bar));
}
```
The output looks like this (done in Linux)
```
$ gcc main.c 
$ ./a.out 
FILE is not defined
sizeof(FILE) 216
```

PR #30, opened 2024-08-22
[noup] zephyr: Set max idle period

ajayparida:
Dynamically set max idle period.

### DIFF
--- a/src/drivers/driver.h
+++ b/src/drivers/driver.h
@@ -1375,6 +1375,9 @@ struct wpa_driver_associate_params {
 	 * rsn_overriding - wpa_supplicant RSN overriding support
 	 */
 	bool rsn_overriding;
+
+	/* max_idle_period - Maximum idle period likely for this connection */
+	u16 max_idle_period;
 };
 
 enum hide_ssid {
--- a/wpa_supplicant/config.c
+++ b/wpa_supplicant/config.c
@@ -4699,6 +4699,7 @@ struct wpa_config * wpa_config_alloc_empty(const char *ctrl_interface,
 	if (driver_param)
 		config->driver_param = os_strdup(driver_param);
 	config->gas_rand_addr_lifetime = DEFAULT_RAND_ADDR_LIFETIME;
+	config->max_idle_period = DEFAULT_MAX_IDLE_PERIOD;
 
 #ifdef CONFIG_TESTING_OPTIONS
 	config->mld_connect_band_pref = DEFAULT_MLD_CONNECT_BAND_PREF;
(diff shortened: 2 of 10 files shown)


krish2718 (changes requested):
This needs to be a proper feature in upstream hostap where `bss_max_idle_period` can be set in multiple ways, config file, wpa_cli etc.

wpa_supplicant/events.c
```diff
@@ -2947,8 +2947,15 @@ static void wnm_process_assoc_resp(struct wpa_supplicant *wpa_s,
...
-			WPA_GET_LE16(elems.bss_max_idle_period);
+
+		if (wpa_s->conf->max_idle_period >=
+			WPA_GET_LE16(elems.bss_max_idle_period)) {
+			wpa_s->sme.bss_max_idle_period =
+				WPA_GET_LE16(elems.bss_max_idle_period);
+		} else {
+			wpa_s->sme.bss_max_idle_period = wpa_s->conf->max_idle_period;
+		}
+
```
krish2718, line 2958: as per the spec, the value in assoc request is just a preference, we should always use the value from the assoc response as is: See `802.11-2024: 11.21.13`

> If the BSS Max Idle Period element is present in the (Re)Association Request frame received by a non-S1G AP that has dot11BSSMaxIdlePeriodIndicationByNonAPSTA equal to true, then the non-S1G AP may choose the non-AP STA’s preferred maximum idle period. The non-S1G AP indicates its chosen value to the non-S1G STA in the (Re)Association Response frame.

wpa_supplicant/config.c
```diff
@@ -5597,6 +5598,7 @@ static const struct global_parse_data global_fields[] = {
 	/* NOTE: When adding new parameters here, add_interface() in
 	 * wpa_supplicant/dbus_new_introspect.c may need to be modified to
 	 * increase the size of the iface->xml buffer. */
```
krish2718, line 5600: what about this comment? I don't see a corresponding change in`wpa_supplicant/dbus_new_introspect.c` is that not needed?

krish2718 (commented):
Where is the code to use the user input to prepare association request frame?

src/drivers/driver.h
```diff
@@ -1375,6 +1375,9 @@ struct wpa_driver_associate_params {
 	 * rsn_overriding - wpa_supplicant RSN overriding support
 	 */
 	bool rsn_overriding;
+
+	/* max_idle_period - Maximum idle period likely for this connection */
+	u16 max_idle_period;
```
krish2718, line 1380: There is already `bss_max_idle_period` here

PR #32, opened 2024-08-30
DPP bug fixes

fengming-ye:
510c2609a8b62acdcd1deaa4e2a9c20bb303f234
[noup] zephyr: fix hostapd_cli set command parse wrong in zephyr
    
    hostapd_config_parse_key_mgmt can parse more than two argcs.
    And hostapd_cli shell command has one more "interactive" argv,
    which makes argc not equal to 2.
    So align with wpa_cli_cmd_set.

a41113eb54fc208b0e22368b9891e13cdba83540
[noup] add hostapd_cli shell support
    
    Add hostapd_cli shell command.
    Porting from zephyr wpa_cli impl.

83d7a2b827515e9c3142fdc629679fb317286be4
[noup] fix hostapd ctrl msg receive build error
    
    Fix build error that 'len' is undefined in hostapd_cli_recv_pending.

960516c8e32a662b73b9aac012d82cc814aa71c4
[noup] hostapd: add dispatch for hostapd and wpa_supplicant wpa_msg
    
    When we have coexistence of hostapd and wpa_supplicant,
    wpa_msg has different implementation and differenet context,
    struct wpa_supplicant and struct hostapd_data.
    So to let them work together, we need to have common implementation
    for wpa_msg and dispatch msgs for hostapd and wpa_supplicant.
    
    In this case add common member is_hostapd in struct wpa_supplicant
    and struct hostapd_data, to judge if wpa_msg context is wpa_supplicant
    or hostapd. In hostapd it is 1.
    And wrap wpa_supplicant msg_cb and hostapd msg_cb by apis called by
    common msg_cb.
    
    So add common member is_hostapd in struct wpa_supplicant
    and struct hostapd_data, to judge if wpa_msg context is wpa_supplicant
    or hostapd. In hostapd it is 1.
    And wrap wpa_supplicant msg_cb and hostapd msg_cb by apis called by
    common msg_cb.

a00f0b566f641f5dffd0da1d16020f8f56bbb29c
[noup] zephyr: fix DPP-Auth initiation failed no response received
    
    Configurator send DPP auth request on channel 1 and need to
    receive DPP auth response on channel 6.
    This requires support of send_action_cancel_wait ops to cancel
    remain on channel after TX success on channel 1, and then
    remain on channel 6.
    But previous code does not support send_action_cancel_wait ops.
    
    So add send_action_cancel_wait ops support to fix this case.

### DIFF
--- a/hostapd/hostapd_cli_zephyr.c
+++ b/hostapd/hostapd_cli_zephyr.c
@@ -263,7 +263,7 @@ static void hostapd_cli_recv_pending(struct wpa_ctrl *ctrl, struct hostapd_data 
 
 			msg->msg[msg->msg_len] = '\0';
 			wpa_printf(MSG_DEBUG, "Received len: %d, msg_len:%d - %s->END\n",
-				   len, msg->msg_len, msg->msg);
+				   plen, msg->msg_len, msg->msg);
 			if (msg->msg_len >= MAX_CTRL_MSG_LEN) {
 				wpa_printf(MSG_INFO, "Too long message received.\n");
 				continue;
--- a/hostapd/hostapd_cli_zephyr.c
+++ b/hostapd/hostapd_cli_zephyr.c
@@ -10,7 +10,6 @@
 #include "includes.h"
 
 #include "common/cli.h"
-#include "common/wpa_ctrl.h"
 #include "utils/common.h"
 #include "utils/eloop.h"
 #include "utils/edit.h"
@@ -211,6 +210,11 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print, 
 int hostapd_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd)
 {
 	return _wpa_ctrl_command(ctrl, cmd, 0, NULL);
+}
+
+int hostapd_ctrl_command_interactive(struct wpa_ctrl *ctrl, const char *cmd)
+{
+	return _wpa_ctrl_command(ctrl, cmd, 1, NULL);
 }
 
 int zephyr_hostapd_cli_cmd_resp(const char *cmd, char *resp)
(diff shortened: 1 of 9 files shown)


jukkar (approved):
LGTM

jukkar:
cc: @krish2718

src/ap/hostapd.h
```diff
@@ -154,6 +154,13 @@ struct hostapd_sae_commit_queue {
  * struct hostapd_data - hostapd per-BSS data structure
  */
 struct hostapd_data {
+#if defined(__ZEPHYR__) && defined(HOSTAPD)
+	/** Zephyr has coexistence of hostapd and wpa_supplicant.
+	 * To verify if ctx is hostapd or wpa_supplicant in wpa_msg,
+	 * this should be the first parameter
+	 */
+	int is_hostapd;
```
krish2718, line 162: As discussed earlier in the hostap ML, there is a clean way to do this (hostap and wpa_s in parallel, using registration of ops), I am accepting this patch on the assumption that we will start work on a proper implementation than these hacks.

hostapd/hostapd_cli_zephyr.c
```diff
@@ -262,8 +262,6 @@ static void hostapd_cli_recv_pending(struct wpa_ctrl *ctrl, struct hostapd_data
 			struct conn_msg *msg = (struct conn_msg *)buf;
 
 			msg->msg[msg->msg_len] = '\0';
-			wpa_printf(MSG_DEBUG, "Received len: %d, msg_len:%d - %s->END\n",
```
krish2718, line 265: is this debug not useful? Should we fix it instead?

src/ap/hostapd.h
fengming-ye: can we go with this temproraily and I'll revert it after you merged proper impl?

krish2718: Sure

krish2718 (approved):

PR #33, opened 2024-08-30
[nrf noup] crypto_alt: Fix secret key len

krish2718:
For EAP-TLS, secret key len is always 48, and MD_SIZE is 32, this causes authentication failure.

### DIFF
--- a/src/crypto/tls_mbedtls_alt.c
+++ b/src/crypto/tls_mbedtls_alt.c
@@ -216,7 +216,7 @@ struct tls_connection
 #if MBEDTLS_VERSION_NUMBER < 0x03000000 /* mbedtls 3.0.0 */
     unsigned char expkey_secret[MBEDTLS_EXPKEY_FIXED_SECRET_LEN];
 #else
-    unsigned char expkey_secret[MBEDTLS_MD_MAX_SIZE];
+    unsigned char expkey_secret[48];
 #endif
     unsigned char expkey_randbytes[MBEDTLS_EXPKEY_RAND_LEN * 2];
 


krish2718:
@fengming-ye @MaochenWang1 FYI

fengming-ye:
/* Note: this should always be >= PSA_HASH_MAX_SIZE
 * in all builds with both CRYPTO_C and MD_LIGHT.
 *
 * This is to make things easier for modules such as TLS that may define a
 * buffer size using MD_MAX_SIZE in a part of the code that's common to PSA
 * and legacy, then assume the buffer's size is PSA_HASH_MAX_SIZE in another
 * part of the code based on PSA.
 */
#if defined(MBEDTLS_MD_CAN_SHA512) || defined(MBEDTLS_MD_CAN_SHA3_512)
#define MBEDTLS_MD_MAX_SIZE         64  /* longest known is SHA512 */
#elif defined(MBEDTLS_MD_CAN_SHA384) || defined(MBEDTLS_MD_CAN_SHA3_384)
#define MBEDTLS_MD_MAX_SIZE         48  /* longest known is SHA384 */
#elif defined(MBEDTLS_MD_CAN_SHA256) || defined(MBEDTLS_MD_CAN_SHA3_256)
#define MBEDTLS_MD_MAX_SIZE         32  /* longest known is SHA256 */
#elif defined(MBEDTLS_MD_CAN_SHA224) || defined(MBEDTLS_MD_CAN_SHA3_224)
#define MBEDTLS_MD_MAX_SIZE         28  /* longest known is SHA224 */
#else
#define MBEDTLS_MD_MAX_SIZE         20  /* longest known is SHA1 or RIPE MD-160
                                           or smaller (MD5 and earlier) */
#endif

MBEDTLS_MD_MAX_SIZE is 48 up to MBEDTLS_SHA384_C. Maybe we should define MBEDTLS_SHA384_C someshere instead of hard code?

#if defined(MBEDTLS_SHA384_C)
#define MBEDTLS_MD_CAN_SHA384
#define MBEDTLS_MD_SOME_LEGACY
#endif

krish2718:
Yeah, but as RFC itself is saying secret key len is 48, I just hardcoded, let me try selecting MD_SHA384 and if it works.

fengming-ye:
> Yeah, but as RFC itself is saying secret key len is 48, I just hardcoded, let me try selecting MD_SHA384 and if it works.

could you share where RFC says it is 48?

fengming-ye:
> MbedTLS itself hardcodes this to 48: https://github.com/Mbed-TLS/mbedtls/blob/ab0af45d11db04bd6e42ccfabe1b8d8deaae3243/library/ssl_tls.c#L9114
> 
> RFC https://datatracker.ietf.org/doc/html/rfc5216#section-2.3 uses PRF-48, so, it's fixed to 48.

In this case we can go with it and see what happens.

jukkar (commented):
Minor nit, the commit subject is wrong as it mentions nrf

krish2718:
> Minor nit, the commit subject is wrong as it mentions nrf

Yes, will fix, though I see `nxp` prefix in some commits, we can clean the tags during planned upmerge.

sachinthegreen (approved):

jukkar (approved):

PR #34, opened 2024-09-02
[noup] zephyr: revert zephyr change for os_time_t to origin

fengming-ye:
Hostap define os_time_t as long but zephyr add own definition to long long and change hostap original code to fix warnings. Thus when DPP enabled, unfixed warning will pop out. We do not want to change so many original hostap code. So it is better to remain original os_time_t definition as long. Fix it by revert os_time_t own definition for zephyr and related printf format change also.

And fix coverity issue by converting time to os_time_t in os_get_time.

### DIFF
--- a/src/utils/os.h
+++ b/src/utils/os.h
@@ -10,13 +10,11 @@
 #define OS_H
 
 #if defined(__ZEPHYR__)
-typedef long long os_time_t;
 #ifndef FILE
 typedef __FILE FILE;
 #endif
-#else
-typedef long os_time_t;
 #endif
+typedef long os_time_t;
 
 /**
  * os_sleep - Sleep (sec, usec)
--- a/src/utils/os_zephyr.c
+++ b/src/utils/os_zephyr.c
@@ -29,8 +29,8 @@ int os_get_time(struct os_time *t)
 	struct timeval tv;
 
 	res = gettimeofday(&tv, NULL);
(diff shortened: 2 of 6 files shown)


krish2718 (commented):
Please divide this into two commits, 1 commit `noup` for revert and another `fromlist` for the original code fix (after submitting to hostap ML).

jukkar:
@fengming-ye any progress on this?

fengming-ye:
> @fengming-ye any progress on this?

splited commit into 2 commits. One reverting previous format change of PRINTF. Another one reverting os_time_t definition change by Zephyr and fix coverity issues.
So all the change in this PR are aligning code with hostap community, except in os_zephyr.c.
So I think they won't go to hostap ML, at least for now.

And could you check downstream whether your coverity issues are fixed?

jukkar (approved):

src/utils/os.h
```diff
@@ -10,13 +10,11 @@
...
 
 #if defined(__ZEPHYR__)
-typedef long long os_time_t;
 #ifndef FILE
 typedef __FILE FILE;
 #endif
-#else
-typedef long os_time_t;
 #endif
+typedef long os_time_t;
```
krish2718, line 17: should this revert also be in the first commit?

jukkar (approved):

PR #37, opened 2024-09-12
[noup] zephyr: crypto: Fix for MbedTLS Enterprise issue

LiLongNXP:
Issue is enterprise sta connect fail after uap start stop.

After uap start stop, tls_ctx_global.tls_conf not freed and still not NULL.
If tls_ctx_global.tls_conf not NULL, it will choose SERVER ROLE when do connect.
STA connect fail for endpoint set to SERVER ROLE
(should be CLIENT ROLE) and respond wrong EAP packet to peer. The hello client packet is without client content and then deauth by AP.
The reason that tls_ctx_global.tls_conf not freed and not set to NULL is for tls_deinit() returned early when tls_ctx_global.refcnt is not 0.
Then it not call tls_conf_deinit() and not set
tls_ctx_global.tls_conf NULL.

Add fix to code to:
Call tls_conf_deinit() before tls_ctx_global.refcnt check. And tls_ctx_global.tls_conf has it's own refcnt.

### DIFF
--- a/src/crypto/tls_mbedtls_alt.c
+++ b/src/crypto/tls_mbedtls_alt.c
@@ -362,10 +362,10 @@ struct tls_conf *tls_conf_init(void *tls_ctx)
     return tls_conf;
 }
 
-void tls_conf_deinit(struct tls_conf *tls_conf)
+struct tls_conf *tls_conf_deinit(struct tls_conf *tls_conf)
 {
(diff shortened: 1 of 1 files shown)


jukkar (approved):
LGTM

krish2718 (approved):

PR #38, opened 2024-09-19
zephyr: Roaming support

nxf58150:
Added roaming support in wpa_supplicant for Zephyr.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -535,6 +535,9 @@ void wpa_drv_zep_event_proc_assoc_resp(struct zep_drv_if_ctx *if_ctx,
 				       unsigned int status)
 {
 	if (status != WLAN_STATUS_SUCCESS) {
+		if (if_ctx->ft_roaming) {
+			if_ctx->ft_roaming = false;
+		}
 		wpa_supplicant_event_wrapper(if_ctx->supp_if_ctx,
 				EVENT_ASSOC_REJECT,
 				event);
@@ -1542,6 +1545,10 @@ static int wpa_drv_zep_deauthenticate(void *priv, const u8 *addr,
 
 	if_ctx = priv;
 
+	if (if_ctx->ft_roaming) {
+		if_ctx->ft_roaming = false;
+	}
+
 	dev_ops = get_dev_ops(if_ctx->dev_ctx);
 	ret = dev_ops->deauthenticate(if_ctx->dev_priv, addr, reason_code);
 	if (ret) {
@@ -1569,6 +1576,12 @@ static int wpa_drv_zep_authenticate(void *priv,
 	}
 
 	if_ctx = priv;
+
+	if_ctx->ft_roaming = false;
+
+	if (params->auth_alg == WPA_AUTH_ALG_FT) {
+		if_ctx->ft_roaming = true;
+	}
 
 	dev_ops = get_dev_ops(if_ctx->dev_ctx);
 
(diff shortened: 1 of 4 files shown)


nxf58150:
The 11kvr roaming RFC link: https://github.com/zephyrproject-rtos/zephyr/issues/76307

src/drivers/driver_zephyr.c
```diff
@@ -530,11 +530,38 @@ void wpa_drv_zep_event_proc_auth_resp(struct zep_drv_if_ctx *if_ctx,
 }
 
 
+static void wpa_drv_zep_free_pairwise_key_params(struct wpa_driver_set_key_params *params)
+{
+	if (params->ifname) {
```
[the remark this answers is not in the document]
nxf58150: Removed all checks. Thanks!

jukkar (approved):

jukkar:
ping @krish2718, please take a look

krish2718 (commented):
Can you please split the 2nd commit in to two? The commit log as bulleted list, a good indication of the need to split the commit.

nxf58150:
> Can you please split the 2nd commit in to two? The commit log as bulleted list, a good indication of the need to split the commit.

Done. Please help review it again. Thanks!

jukkar:
I think this roaming PR could go next, @krish2718 please take a look of this and also the zephyr counter part.

src/drivers/driver_zephyr.c
```diff
@@ -1720,6 +1817,29 @@ static int _wpa_drv_zep_set_key(void *priv,
 static int wpa_drv_zep_set_key(void* priv,
 			       struct wpa_driver_set_key_params *params)
 {
+	struct zep_drv_if_ctx *if_ctx = NULL;
+	enum key_flag key_flag        = params->key_flag;
+
+	if_ctx = priv;
+
+	if (if_ctx->ft_roaming && (key_flag & KEY_FLAG_PAIRWISE)) {
+		return wpa_drv_zep_save_pairwise_key_params(priv, params);
```
krish2718, line 1826: The issue itself looks generic enough but this fix is a hack, without using extended keys (storing both keys in parallel) the packet loss is common either in re-keying or roaming, do we know how the nl80211 handles this scenario? I can also check that and see if we can come up with a proper solution than storing and installing keys at a later point.

nxf58150: I removed the fix in zephyr driver and added it to our wifi driver. If other vendor uses zephyr also has this issue, we can discuss more generic fix in zephyr driver in the future. Thanks!

MaochenWang1:
hi @krish2718  please review this, we need this PR to make the upstream NXP wifi example build pass.

src/drivers/driver_zephyr.c
```diff
@@ -1812,7 +1832,11 @@ static int wpa_drv_zep_set_supp_port(void *priv,
 
 #ifdef CONFIG_NET_DHCPV4
 	if (authorized) {
-		net_dhcpv4_restart(iface);
+		if (if_ctx->ft_roaming == false) {
+			net_dhcpv4_restart(iface);
+		} else {
+			if_ctx->ft_roaming = false;
+		}
```
krish2718, line 1839: This only works for L2 roaming, should this be a Kconfig option? or an alternative solution?

nxf58150: What do you mean by Kconfig option?

krish2718: Something like below? Also please extend the commit log to explain

```
config WIFI_NM_WPA_SUPPLICANT_SKIP_DHCP_ON_ROAMING
   bool "Do not restart DHCP when roaming e.g., L2 roaming"
   help
     For L2 roaming, the IP network doesn't change, so, DHCP can be skipped.
     For L3 roaming this needs to be enabled.
```

nxf58150: Could you please help clarify L2 roaming and L3 roaming? Thanks!

krish2718: L2 romaing, romaing within same subnet, L3 roaming is roaming across subnets, some info https://learningnetwork.cisco.com/s/question/0D56e0000CjUe4lCQC/what-is-layer2-and-layer3-roaming . The point is this decision of whether to perform DHCP renewal or not should be user configurable as each scenario might need a different behaviour.

nxf58150: Ok, got it. I will update the PR later. Thanks!

nxf58150: Hi @krish2718 , I added new kconfig WIFI_NM_WPA_SUPPLICANT_SKIP_DHCP_ON_ROAMING as suggested and wrapped the source code in this new config. Please help review it. Thanks!

PR #39, opened 2024-09-19
[noup] Fix incorrect length increment

krish2718:
If the incoming cert is not a PEM then the length is incremented to handle null byte, but null byte itself is not added, this causes MbedTLS to parse the extra byte and causing UNKNOWN_TAG parser error depending on the value at that byte.

In case the user doesn't give PEM then let MbedTLS handle parsing and failing appropriately. MbedTLS has much stricter PEM format checks than this file.


krish2718:
@fengming-ye FYI

jukkar (approved):

sachinthegreen (approved):

fengming-ye (approved):

fengming-ye:
@MaochenWang1 FYI

PR #40, opened 2024-09-25
[noup] zephyr: crypto: remove unused crypto apis

fengming-ye:
Remove crypto apis not found caller in hostap lib.


jukkar (approved):
I am wondering why these not used APIs were introduced at all when the file was imported by you. Anyway, if these are not needed, then lets remove them.

fengming-ye:
> I am wondering why these not used APIs were introduced at all when the file was imported by you. Anyway, if these are not needed, then lets remove them.

Because they are implemented in other crypto wrappers but not seen in hostap, which I realized now.

krish2718 (approved):

PR #43, opened 2024-09-25
[noup] base64: Rename base64_encode/decode functions

rlubos:
Zephyr provides own base64 library with the same function names as hostap. Because of this, it's not possible to build applications enabling both supplicant and CONFIG_BASE64. As Zephyr takes preference here, we need to rename the colliding functions in hostap.

### DIFF
--- a/hostapd/sae_pk_gen.c
+++ b/hostapd/sae_pk_gen.c
@@ -131,7 +131,7 @@ int main(int argc, char *argv[])
 	fprintf(stderr, "\nFound a valid hash in %llu iterations: %s\n",
 		i + 1, hash_hex);
 
-	b64 = base64_encode(der, der_len, NULL);
+	b64 = hostap_base64_encode(der, der_len, NULL);
 	if (!b64)
 		goto fail;
 	src = pos = b64;
--- a/hs20/client/est.c
+++ b/hs20/client/est.c
@@ -158,7 +158,7 @@ int est_load_cacerts(struct hs20_osu_client *ctx, const char *url)
 		return -1;
 	}
 
-	pkcs7 = base64_decode(resp, resp_len, &pkcs7_len);
+	pkcs7 = hostap_base64_decode(resp, resp_len, &pkcs7_len);
 	if (pkcs7 && pkcs7_len < resp_len / 2) {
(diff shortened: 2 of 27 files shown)


rlubos:
An alternative to this would be removing `base64.c` from build as we did in NCS:
https://github.com/nrfconnect/sdk-nrf/commit/213fb11babf14d06cc76e7fd6bd9755a55614d03

I'm not sure though if this workaround is still viable given that we're aiming for enterprise support in Zephyr. Some of the conditionally compiled source files seem to be `base64_encode/decode()` users.

@MaochenWang1 @Rex-Chen-NXP @fengming-ye Can you check if the Enterprise mode support requires `base64.c` from hostap?

fengming-ye:
- Can you check if the Enterprise mode support requires base64.c from hostap?

you mean base64.c is not required in basic wpa_supplicant & soft AP right?
But base64.c is required in DPP, WPS and enterprise.

But this PR will be really hard to upstream as it is Zephyr special requirement.
So I''d prefer alternative option to remove base64.c and add __ZEPHYR__ macro in base64.h like
#ifdef __ZEPHYR__
#include "sys/base64.h"
#else
char * base64_encode(const void *src, size_t len, size_t *out_len);
#endif

jukkar:
> So I''d prefer alternative option to remove base64.c and add **ZEPHYR** macro in base64.h like
> #if defined(**ZEPHYR** ) && defined(CONFIG_BASE64)
> #include "sys/base64.h"
> #else
> char * base64_encode(const void *src, size_t len, size_t *out_len);
> #endif

Problem is that the base64_encode/decode function prototypes are different in hostap vs zephyr, modifications are needed in hostap side when calling the functions.

fengming-ye:
> > So I''d prefer alternative option to remove base64.c and add **ZEPHYR** macro in base64.h like
> > #if defined(**ZEPHYR** ) && defined(CONFIG_BASE64)
> > #include "sys/base64.h"
> > #else
> > char * base64_encode(const void *src, size_t len, size_t *out_len);
> > #endif
> 
> Problem is that the base64_encode/decode function prototypes are different in hostap vs zephyr, modifications are needed in hostap side when calling the functions.

Okay in this case and base64 is mandatory. I agree with this PR.

jukkar (approved):

PR #44, opened 2024-09-26
[toup] zephyr: crypto: Add PRF-256 Wrapper based on PSA APIs

VivekUppunda:
This commit adds PRF-256 wrapper API for section 11.6.1.7.2 in 802.11ac using PSA APIs. The wrapper API has been validated via test vectors derived during connection and has provided the desired output.

### DIFF
--- a/src/crypto/crypto_mbedtls_alt.c
+++ b/src/crypto/crypto_mbedtls_alt.c
@@ -605,10 +605,57 @@ int sha384_prf(
     return hmac_prf_bits(key, key_len, label, data, data_len, buf, buf_len * 8, MBEDTLS_MD_SHA384);
 }
 
+/**
+ * Based on Supplicant internal implementaion of SHA-256. This API
+ * uses PSA APIs instead of Supplicant internal implementation or
+ * mbedtls APIs.
+ */
+static int hmac_prf256(const u8 *key,
+                       size_t key_len,
+                       const char *label,
+                       const u8 *data,
+                       size_t data_len,
+                       u8 *buf_in,
+                       size_t buf_len_bits,
+                       mbedtls_md_type_t md_type)
+{
+    unsigned short ctr, n_le = host_to_le16(buf_len_bits);
+    const u8 *addr[]         = {(u8 *)&ctr, (u8 *)label, data, (u8 *)&n_le};
+    const size_t len[]       = {2, os_strlen(label), data_len, 2};
+    size_t buf_len           = (buf_len_bits + 7) / 8;
+    u8 *buf  = buf_in;
+
+    for (ctr = 1; buf_len >= SHA256_MAC_LEN; buf_len -= SHA256_MAC_LEN, ++ctr)
+    {
+            if (hmac_sha256_vector(key, key_len, 4, addr, len, buf))
+                    return -1;
+            buf += SHA256_MAC_LEN;
+    }
+
+    if (buf_len)
+    {
+            u8 hash[SHA256_MAC_LEN];
+            if (hmac_sha256_vector(key, key_len, 4, addr, len, hash))
+                    return -1;
+            os_memcpy(buf, hash, buf_len);
+            forced_memzero(hash, sizeof(hash));
+    }
+
+    /* Mask out unused bits in last octet if it does not use all the bits */
+    if ((buf_len_bits &= 0x7))
+            buf[-1] &= (u8)(0xff << (8 - buf_len_bits));
+
+    return 0;
+}
+
 int sha256_prf(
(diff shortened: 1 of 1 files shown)


VivekUppunda:
@fengming-ye - Hi, can you kindly review this Wrapper API? This is only for PRF-256. It is not generic as hmac_prf_bits.

fengming-ye:
Code LGTM. But why don't we add PSA wrapper to hmac_prf_bits?

VivekUppunda:
> Code LGTM. But why don't we add PSA wrapper to hmac_prf_bits?

Presently, I am unable to get test vectors for SHA-384 and SHA-512. If that were available, we could extend this API wrapper to support the same or write wrappers for SHA-384/SHA-512 similar to this API. 

Thanks and regards
Vivek

fengming-ye:
> > Code LGTM. But why don't we add PSA wrapper to hmac_prf_bits?
> 
> Presently, I am unable to get test vectors for SHA-384 and SHA-512. If that were available, we could extend this API wrapper to support the same or write wrappers for SHA-384/SHA-512 similar to this API.
> 
> Thanks and regards Vivek

Okay and could you pls add test case in commit msg so that we can test it also?

VivekUppunda:
> > > Code LGTM. But why don't we add PSA wrapper to hmac_prf_bits?
> > 
> > 
> > Presently, I am unable to get test vectors for SHA-384 and SHA-512. If that were available, we could extend this API wrapper to support the same or write wrappers for SHA-384/SHA-512 similar to this API.
> > Thanks and regards Vivek
> 
> Okay and could you pls add test case in commit msg so that we can test it also?

We have tested this against WPA2-PSK-256 profiles and the test cases for those using this API have been known to have passed. I have updated the commit message as well

src/crypto/crypto_mbedtls_alt.c
```diff
@@ -605,10 +605,57 @@ int sha384_prf(
...
+{
+    unsigned short ctr, n_le = host_to_le16(buf_len_bits);
+    const u8 *addr[]         = {(u8 *)&ctr, (u8 *)label, data, (u8 *)&n_le};
+    const size_t len[]       = {2, os_strlen(label), data_len, 2};
+    size_t buf_len           = (buf_len_bits + 7) / 8;
+    u8 *buf  = buf_in;
+
+    for (ctr = 1; buf_len >= SHA256_MAC_LEN; buf_len -= SHA256_MAC_LEN, ++ctr)
+    {
+            if (hmac_sha256_vector(key, key_len, 4, addr, len, buf))
```
fengming-ye, line 630: this seems not using addr[i] as input.

src/crypto/crypto_mbedtls_alt.c
```diff
@@ -605,10 +605,57 @@ int sha384_prf(
...
+                    return -1;
+            buf += SHA256_MAC_LEN;
+    }
+
+    if (buf_len)
+    {
+            u8 hash[SHA256_MAC_LEN];
+            if (hmac_sha256_vector(key, key_len, 4, addr, len, hash))
```
fengming-ye, line 638: here also

fengming-ye:
I think the best test is, with the all same input (which can be dumped by debug log), the old version (hmac_prf_bits) and the new version (hmac_sha256_prf) have the same output.
Can you add this only in comments?
Test can refer to crypto_module_tests test_sha256().

src/crypto/crypto_mbedtls_alt.c
VivekUppunda: Kindly refer `sha1_t_prf` or `sha1_prf` in the same file. Also, the internal implementation of sha256 PRF. when the hmac API is called, the number of elements is passed as a parameter to the relevant API. In this case which is 4 elements in the array addr. I have followed the same. The output for the WPA2-PSK-256 connection passed and the test vectors tested also provided desired result

src/crypto/crypto_mbedtls_alt.c
VivekUppunda: Same comment as above comment

VivekUppunda:
> I think the best test is, with the all same input (which can be dumped by debug log), the old version (hmac_prf_bits) and the new version (hmac_sha256_prf) have the same output. Can you add this only in comments? Test can refer to crypto_module_tests test_sha256().

we have collected the test vectors via debug log dump in hmac_prf_bits and verified the same with the new implementation. I will update the commit log as well.

fengming-ye (approved):

krish2718 (approved):

jukkar (approved):

PR #46, opened 2024-09-29
zephyr: udp throughput fail when legacy roaming

GaofengZhangNXP:
udp tx throughput fail is expected, legacy roaming triggered dhcp will change ip address, but udp tx throughput specify fixed ip address during udp tx iperf, so udp tx throughput fail after dhcp. Using assoc_freq in wps_supplicant to get current connect state. if assoc_freq is not zero, sta is connected, and we think it is in roaming when auth. doesn't restart dhcp. and ecsa also doesn't restart dhcp.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -541,6 +541,9 @@ void wpa_drv_zep_event_proc_assoc_resp(struct zep_drv_if_ctx *if_ctx,
 		wpa_supplicant_event_wrapper(if_ctx->supp_if_ctx,
 				EVENT_ASSOC_REJECT,
 				event);
+		if (if_ctx->roaming) {
+		    if_ctx->roaming = false;
+		}
 	} else {
 		if_ctx->associated = true;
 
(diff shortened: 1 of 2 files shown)


jukkar:
Please no merge commits.
The "fix print too long msg" is from #50 so can be removed from this PR
Please add "zephyr: " prefix to the commit subject so that it is easier to detect zephyr specific commits.

GaofengZhangNXP:
> Please no merge commits.

update

GaofengZhangNXP:
> Please no merge commits. The "fix print too long msg" is from #50 so can be removed from this PR Please add "zephyr: " prefix to the commit subject so that it is easier to detect zephyr specific commits.

update

src/drivers/driver_zephyr.c
```diff
@@ -1812,7 +1823,11 @@ static int wpa_drv_zep_set_supp_port(void *priv,
 
 #ifdef CONFIG_NET_DHCPV4
 	if (authorized) {
-		net_dhcpv4_restart(iface);
+		if (if_ctx->roaming == true) {
+			if_ctx->roaming = false;
```
krish2718, line 1827: This doesn't work for L3 roaming where DHCP is necessary, the proper fix would be for the application to implement NET_MGMT IPV4 events and then restart UDP throughput (zperf?).

PR #48, opened 2024-10-06
sync fixes from NCS

krish2718:
Now that NCS uses upstream hostap, but few fixes are missed, so, pull them to upstream.


rlubos (approved):

sachinthegreen (approved):

jukkar:
The changes are ok, but the the commit messages contain Nordic specific information which we could cleanup.

krish2718:
> The changes are ok, but the the commit messages contain Nordic specific information which we could cleanup.

Ah I just cherry-picked NCS commits, I can clean them up.

jukkar (approved):

PR #49, opened 2024-10-09
[no up] ap: Fix double event sending

krish2718:
When AP is enabled using WPA supplicant we get two AP enable events, one from WPA supplicant and other from hostapd.

FYI @nxf58150


nxf58150:
> When AP is enabled using WPA supplicant we get two AP enable events, one from WPA supplicant and other from hostapd.
> 
> FYI @nxf58150

Got it. Thanks!

jukkar (approved):

rlubos (approved):
nit: The tag could be `[noup]` as others (no space).

PR #51, opened 2024-10-10
zephyr: cert: fix check AP certificate file fail

fengming-ye:
For some certificate files may check key usage fail and send alert message to AP.
Enable TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK to disable key usage check.

### DIFF
--- a/src/crypto/tls_mbedtls_alt.c
+++ b/src/crypto/tls_mbedtls_alt.c
@@ -121,7 +121,9 @@ typedef mbedtls_tls_prf_types int;
  * then checking the client cert against the DN hint in certificate verify_cb.
  * Comment out to disable feature and remove ~3k from binary .text segment */
 #define TLS_MBEDTLS_CERT_VERIFY_EXTMATCH
-//#define TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK
+#ifdef MBEDTLS_ALLOW_PRIVATE_ACCESS
+#define TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK
+#endif
 
 #if defined(EAP_FAST) || defined(EAP_FAST_DYNAMIC) || defined(EAP_SERVER_FAST) || defined(EAP_TEAP) || \
     defined(EAP_SERVER_TEAP)


src/crypto/tls_mbedtls_alt.c
```diff
@@ -121,7 +121,9 @@ typedef mbedtls_tls_prf_types int;
  * then checking the client cert against the DN hint in certificate verify_cb.
  * Comment out to disable feature and remove ~3k from binary .text segment */
 #define TLS_MBEDTLS_CERT_VERIFY_EXTMATCH
-//#define TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK
+#ifdef MBEDTLS_ALLOW_PRIVATE_ACCESS
+#define TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK
```
krish2718, line 125: Can you please share a reference to this define? I couldn't find it anywhere, is this NXP internal? And also how is it reliant on `MBEDTLS_ALLOW_PRIVATE_ACCESS`? As per docs, the private access using this is deprecated and if you still need access then this has to be raised upstream.

fengming-ye: TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK is used inside this file to disable key usage check, which is private access.

krish2718: Ah sorry missed it, I did a search and couldn't find it but was on the wrong branch, thanks.

krish2718: Can you please also elaborate on how does the flags `MBEDTLS_X509_EXT_KEY_USAGE` and `MBEDTLS_X509_EXT_EXTENDED_KEY_USAGE` tie in to private access, I see that `key_usage` is private in  `mbedtls_x509_crt` but don't  see that used in this file? and how does the reporting to AP work?

krish2718: Thanks for the clarification, one last question, why the dependency on `MBEDTLS_ALLOW_PRIVATE_ACCESS`? the `key_usage` is marked as private in `mbedtls_x509_crt` but unused by the hostap.

fengming-ye: tls_mbedtls_verify_cb for TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK,
it disables MBEDTLS_X509_EXT_KEY_USAGE and MBEDTLS_X509_EXT_EXTENDED_KEY_USAGE of ext_types,
to disable key usage check, which is MBEDTLS_PRIVATE.
So for applications not defining MBEDTLS_ALLOW_PRIVATE_ACCESS, we don't enable TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK to avoid build error.

krish2718:
```
#if defined(TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK)
    crt->ext_types &= ~MBEDTLS_X509_EXT_KEY_USAGE;
    crt->ext_types &= ~MBEDTLS_X509_EXT_EXTENDED_KEY_USAGE;
#endif
```
I only see above changes and both are enumerations, isn't private access limited to structure members?

fengming-ye: Hi @krish2718 any comments for this PR?

krish2718 (approved):

jukkar (approved):

PR #54, opened 2024-10-16
[noup] mbedtls: Modify MBEDTLS_NIST_KW_C to avoid name conflict

rado17:
Same CFLAGS are being used in MbedTLS and Hostap for MBEDTLS_NIST_KW_C which causes build time warnings. Change the CFLAG in hostap to avoid this conflict.


krish2718 (changes requested):
This looks incorrect, the MbedTLS macro should be same else this code will be compiled out.

rado17:
> This looks incorrect, the MbedTLS macro should be same else this code will be compiled out.

Understood. Closing this PR.

PR #56, opened 2024-10-17
[toup] crypto: remove the include of fsl_debug_console.h

MaochenWang1:
Remove the include of fsl_debug_console.h, which is not used for Zephyr

### DIFF
--- a/src/crypto/tls_mbedtls_alt.c
+++ b/src/crypto/tls_mbedtls_alt.c
@@ -64,8 +64,6 @@
 #include <mbedtls/debug.h>
 #ifdef __ZEPHYR__
 #define PRINTF printk
-#else
-#include "fsl_debug_console.h"
 #endif
 #define tls_mbedtls_d(...) PRINTF("tls_mbedtls", ##__VA_ARGS__)
 #else


src/crypto/tls_mbedtls_alt.c
```diff
@@ -64,8 +64,6 @@
 #include <mbedtls/debug.h>
 #ifdef __ZEPHYR__
 #define PRINTF printk
-#else
-#include "fsl_debug_console.h"
```
krish2718, line 68: isn't this for non-zephyr? commit log says it's unused in Zephyr?

MaochenWang1: "fsl_debug_console.h" is a NXP header file, should be removed here.

krish2718: I got that, but why was it used for non-zephyr branch, was it intentional or mistake? Just trying to understand the original change.

MaochenWang1: yes, it was used for non-zephyr branch and it's intentional to support printf.
but adding here is mistake.

krish2718 (approved):

jukkar (approved):

PR #57, opened 2024-10-18
Fix/compile errors or warnings

jukkar:
Something I noticed while working on build tests for hostap in https://github.com/zephyrproject-rtos/zephyr/pull/79973

### DIFF
--- a/src/common/dpp_tcp.c
+++ b/src/common/dpp_tcp.c
@@ -583,7 +583,7 @@ int dpp_relay_rx_gas_req(struct dpp_global *dpp, const u8 *src, const u8 *data,
 			 size_t data_len)
 {
 	struct dpp_relay_controller *ctrl;
-	struct dpp_connection *conn, *found = NULL;
+	struct dpp_connection *conn = NULL, *found = NULL;
 	struct wpabuf *msg;
 
 	/* Check if there is a successfully completed authentication for this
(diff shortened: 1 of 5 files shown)


rlubos (approved):

src/common/dpp_tcp.c
```diff
@@ -583,7 +583,7 @@ int dpp_relay_rx_gas_req(struct dpp_global *dpp, const u8 *src, const u8 *data,
 			 size_t data_len)
 {
 	struct dpp_relay_controller *ctrl;
-	struct dpp_connection *conn, *found = NULL;
+	struct dpp_connection *conn = NULL, *found = NULL;
```
dleach02, line 586: nice... this is why I also don't like doing this... my personal style is to put each declare on its own line.

wpa_supplicant/p2p_supplicant.c
```diff
@@ -5210,7 +5210,7 @@ static int wpas_p2p_peer_go(struct wpa_supplicant *wpa_s,
 	updated = os_reltime_before(&wpa_s->p2p_auto_started,
 				    &bss->last_update);
 	wpa_printf(MSG_DEBUG, "P2P: Current BSS entry for peer updated at "
-		   "%lld.%06lld (%supdated in last scan)",
+		   "%ld.%06ld (%supdated in last scan)",
```
dleach02, line 5213: is this format architecture dependent? I'm guessing this is a warning if you do it wrong?

dleach02 (approved):

wpa_supplicant/p2p_supplicant.c
jukkar: This is actually reverting the changes I did when porting hostap to upstream zephyr, so I added the `%lld`, but then dbe9b099edf2c54ce2c04c93ac6b952dbb798fe9 changed the `os_time_t` back to the original (in order to not do so much changes in the code). But some of my conversions were missed.

jukkar:
@krish2718 PTAL

krish2718 (approved):

PR #58, opened 2024-10-19
[noup] wpa_ctrl: Add a config option for WPA control timeout

krish2718:
WPA control interface timeout is hardcoded to 10s, use the newly introduced configuration option, this is needed sometimes as a workaround e.g., crypto taking too long to complete the request.

Work around for #79834.

Note: Though a generic change, this is marked as `noup` because typically 10s timeout is plenty and I don't think an extra config option would be welcomed :).

### DIFF
--- a/src/common/wpa_ctrl.c
+++ b/src/common/wpa_ctrl.c
@@ -548,7 +548,7 @@ retry_send:
 	os_free(cmd_buf);
 
 	for (;;) {
-		tv.tv_sec = 10;
+		tv.tv_sec = CONFIG_WIFI_NM_WPA_CTRL_RESP_TIMEOUT_S;
 		tv.tv_usec = 0;
 		FD_ZERO(&rfds);
 		FD_SET(ctrl->s, &rfds);


krish2718:
@fengming-ye FYI

fengming-ye:
Okay to me. Though it does not fix our issue of WPA2 AP enable slowly. And I must say if some response comes after 10 seconds, it may be an issue. So we need to find the root cause. This may cover some slow reponse by our change or origin wpa_supplicant impl(which I doubt would happen).
But agian I'm okay with this.

fengming-ye (approved):

sachinthegreen (approved):

krish2718:
>  And I must say if some response comes after 10 seconds, it may be an issue. So we need to find the root cause. This may cover some slow reponse by our change or origin wpa_supplicant impl(which I doubt would happen).

Absolutely this is just a temporary workaround to increase from 10s -> 15s, I will revert this once we find the root cause of why PBKDF2 takes longer (and time varies per-sample).

jukkar (approved):

PR #59, opened 2024-10-22
Fix mbedtls private access

krish2718:
Fixes a access of private members.

@fengming-ye FYI.

### DIFF
--- a/src/crypto/tls_mbedtls_alt.c
+++ b/src/crypto/tls_mbedtls_alt.c
@@ -76,10 +76,6 @@
 #include <mbedtls/net.h>
 #endif
 
-#ifndef MBEDTLS_PRIVATE
-#define MBEDTLS_PRIVATE(x) x
-#endif
-
 #if MBEDTLS_VERSION_NUMBER < 0x03020000 /* mbedtls 3.2.0 */
 #define mbedtls_ssl_get_ciphersuite_id_from_ssl(ssl) \
     ((ssl)->MBEDTLS_PRIVATE(session) ? (ssl)->MBEDTLS_PRIVATE(session)->MBEDTLS_PRIVATE(ciphersuite) : 0)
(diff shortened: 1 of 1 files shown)


fengming-ye (approved):

fengming-ye:
Thanks for the fix. But I still wonder why this build error happens. For tls_mbedtls_alt.c if TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK defined, MBEDTLS_ALLOW_PRIVATE_ACCESS must be defined. Then in private_access.h #define MBEDTLS_PRIVATE(member) member. So tls_mbedtls_alt.c will include x509_crt.h of struct mbedtls_x509_crt -> int MBEDTLS_PRIVATE(ext_types) as ->ext_types.
So why is there build error?

rlubos (approved):

krish2718:
> Thanks for the fix. But I still wonder why this build error happens. For tls_mbedtls_alt.c if TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK defined, MBEDTLS_ALLOW_PRIVATE_ACCESS must be defined. Then in private_access.h #define MBEDTLS_PRIVATE(member) member. So tls_mbedtls_alt.c will include x509_crt.h of struct mbedtls_x509_crt -> int MBEDTLS_PRIVATE(ext_types) as ->ext_types. So why is there build error?

As I said in the commit log, there was a mismatch where `x509_crt.h` was declaring it as private. In our case below is the scenario

```
#include <mbedtls/private_access.h> // as MBEDTLS_ALLOW_PRIVATE_ACCESS is not enabled, the macro is `private_##member`
#include <mbedtls/x509_crt.h> // member is `private_ext_types`
#include <some_other_platform_file> // this file does: #define MBEDTLS_ALLOW_PRIVATE_ACCESS

#ifdef MBEDTLS_ALLOW_PRIVATE_ACCESS
#define TLS_MBEDTLS_CERT_DISABLE_KEY_USAGE_CHECK // this will now be enabled due to above define
#endif

crt->ext_types; // this fails

MBEDTLS_PRIVATE(ext_types); // this works because, the macros is already instantiated with private_##member.
```

jukkar (approved):

PR #60, opened 2024-10-23
Fix unused variable warning

krish2718:
Build error when debug is disabled.

### DIFF
--- a/wpa_supplicant/wpa_supplicant.c
+++ b/wpa_supplicant/wpa_supplicant.c
@@ -8229,9 +8229,11 @@ void dump_freq_data(struct wpa_supplicant *wpa_s, const char *title,
 	wpa_dbg(wpa_s, MSG_DEBUG, "Shared frequencies (len=%u): %s",
 		len, title);
 	for (i = 0; i < len; i++) {
+#if CONFIG_WIFI_NM_WPA_SUPPLICANT_DEBUG_LEVEL <= MSG_DEBUG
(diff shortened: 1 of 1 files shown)


rlubos (approved):

wpa_supplicant/wpa_supplicant.c
```diff
@@ -8229,9 +8229,11 @@ void dump_freq_data(struct wpa_supplicant *wpa_s, const char *title,
 	wpa_dbg(wpa_s, MSG_DEBUG, "Shared frequencies (len=%u): %s",
 		len, title);
 	for (i = 0; i < len; i++) {
+#if CONFIG_WIFI_NM_WPA_SUPPLICANT_DEBUG_LEVEL <= MSG_DEBUG
```
jukkar, line 8232: Hmm, the config option is only available in zephyr, isn't there a config option that is set also in supplicant upstream code and in zephyr for debug level?

jukkar: Or could we do similar things as what `ARG_UNUSED()` does?

krish2718: > isn't there a config option that is set also in supplicant upstream code and 

No, and even if it did the code doesn't get compiled out, so, this warning won't be seen. This was an memory optimization done by Nordic.

jukkar (approved):

PR #61, opened 2024-10-25
zephyr: udp throughput fail when legacy roaming

GaofengZhangNXP:
zperf udp tx throughput uses fixed ip address, if run legacy roaming at
this time, roaming will restart dhcp and ip address is changed,
so zperf can't find ip address when zperf send udp packet during legacy
roaming, and zperf udp tx fail.
we think that legacy roaming shouldn't restart dhcp, so we need to
determine whether current connection is a normal connection or a
roaming, and determine whether we should restart dhcp.
and we use assoc_freq parameter in wps_supplicant to get current connect
state.
if assoc_freq is not 0, it indicates that the sta already connects, and
current connection is roaming, so we doesn't start dhcp.
if assoc_freq is 0, it indicates that current connection is a normal
connection rather than roaming, so we want to start dhcp.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -538,6 +538,9 @@ void wpa_drv_zep_event_proc_assoc_resp(struct zep_drv_if_ctx *if_ctx,
 		if (if_ctx->ft_roaming) {
 			if_ctx->ft_roaming = false;
 		}
+		if (if_ctx->roaming) {
+		    if_ctx->roaming = false;
+		}
 		wpa_supplicant_event_wrapper(if_ctx->supp_if_ctx,
 				EVENT_ASSOC_REJECT,
 				event);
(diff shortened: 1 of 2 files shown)


krish2718:
Please raise a manifest PR too.

GaofengZhangNXP:
> Please raise a manifest PR too.
manifest PR: 
https://github.com/zephyrproject-rtos/zephyr/pull/82587

src/drivers/driver_zephyr.c
```diff
@@ -538,6 +538,9 @@ void wpa_drv_zep_event_proc_assoc_resp(struct zep_drv_if_ctx *if_ctx,
 		if (if_ctx->ft_roaming) {
```
jukkar, line 538: @GaofengZhangNXP please re-write the commit message with proper English, it is very difficult to understand what it is saying. Also please start the sentences with capital letter so that it is easier to separate the sentences.
So the commit message should say the reasoning why this change is being done, this is not clear from the message.

GaofengZhangNXP: @jukkar I have updated the commit, please help review it, thanks

krish2718 (approved):

GaofengZhangNXP:
Hi @jukkar , please help review the patch, if you have not any other suggestion, please help merge it, thank you very much.

gannicus-gj:
Hi @jukkar 
could you pls take time to review before Christmas holiday?

jukkar (approved):

PR #62, opened 2024-10-25
A few trivial fixes to build warnings on 64bit targets

aescolar:
I have no clue how you guys are handling this fork, if you want me to add some tag to the commit titles or so just tell.
You are also welcome to modify the commits or do anything with them as you please.

    Fix multiple size_t print format specifiers
    
    The format specifier for size_t is zu.
    Using d only works when int and size_t are the same
    underlying type which is not the case for 64bit systems,
    which leads to a build warning in this case.
    
    In 2 places, instead of changing the printf specifier
    the type of the variable was changed to int, as that
    was what the code really exepected.
    And while at it, two internal function have been made static.

----

    Fix overflow warning for 64bit builds
    
    Fix the following build warning:
    
    ap_drv_ops.c:272:41: warning: overflow in conversion from
    ‘long unsigned int’ to ‘int’ changes value from ‘18446744073709551614’
    to ‘-2’ [-Woverflow]
      272 |                                      0, ~WPA_STA_AUTHORIZED);
    

Partial fix for
https://github.com/zephyrproject-rtos/zephyr/issues/80437

### DIFF
--- a/hostapd/hostapd_cli_zephyr.c
+++ b/hostapd/hostapd_cli_zephyr.c
@@ -126,7 +126,7 @@ static char make_argv(char **ppcmd, uint8_t c)
 	return quote;
 }
 
-char hostapd_make_argv(size_t *argc, const char **argv, char *cmd,
+static char hostapd_make_argv(int *argc, const char **argv, char *cmd,
 		       uint8_t max_argc)
 {
 	char quote = 0;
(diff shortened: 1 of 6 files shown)


jukkar:
Hmm, I think at this point we could avoid compiling with native_sim64. We do not want to deviate too much from upstream with this kind of changes.

aescolar:
> Hmm, I think at this point we could avoid compiling with native_sim64. We do not want to deviate too much from upstream with this kind of changes.

It's fine if you want to handle it that way. The issues, even if trivial, are valid though, so I guess we can also queue it to the upstream.

jukkar:
> The issues, even if trivial, are valid though, so I guess we can also queue it to the upstream.

Yes that is true. We have a plan to merge upstream changes after 4.0 so this might be fixed already.

aescolar:
Closing this PR assuming there is no interest to merge it.

krish2718:
> > The issues, even if trivial, are valid though, so I guess we can also queue it to the upstream.
> 
> Yes that is true. We have a plan to merge upstream changes after 4.0 so this might be fixed already.

Except for couple of files rest are in Zephyr specific code, so, may be if you can post the fix for those two files upstream then we can accept this (but in two separate commits, upstream and zephyr).

PR #63, opened 2024-10-28
zephyr: cert: Add blob element to support parse cert data by blob

Rex-Chen-NXP:
As we do not have file system support, add cert blob element
to allow parse cert data by blob instead of file for SAP.

### DIFF
--- a/src/ap/ap_config.h
+++ b/src/ap/ap_config.h
@@ -407,9 +407,15 @@ struct hostapd_bss_config {
 	int ctrl_interface_gid_set;
 
 	char *ca_cert;
+	const u8 *ca_cert_blob;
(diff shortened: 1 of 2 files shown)


jukkar:
@Rex-Chen-NXP could you fix the commit message to
```
As we do not have file system support, add cert blob element
to allow parse cert data by blob instead of file for SAP.
```
as the original is hard to understand.

Rex-Chen-NXP:
> @Rex-Chen-NXP could you fix the commit message to
> 
> ```
> As we do not have file system support, add cert blob element
> to allow parse cert data by blob instead of file for SAP.
> ```
> 
> as the original is hard to understand.

Yes, of course, updated as comment.

PR #64, opened 2024-11-08
Fix ECSA warning log and current_mode is not correct.

MaochenWang1:
Warning prints "<wrn> wpa_supp: Failed to check if DFS is required; ret=-1" are seen during SAP switch to 2.4G's channel using ECSA CMD when SAP started on 5G channels first. The root cause is that when SAP switch to 2.4G, its iface->current_mode is still about 5G, not updated. Check the upstream hostap code, there is already related fix, so porting it here.
fix: https://github.com/zephyrproject-rtos/zephyr/issues/81123

### DIFF
--- a/src/ap/drv_callbacks.c
+++ b/src/ap/drv_callbacks.c
@@ -911,6 +911,12 @@ void hostapd_event_ch_switch(struct hostapd_data *hapd, int freq, int ht,
 		break;
 	}
 
+	/* The operating channel changed when CSA finished, so need to update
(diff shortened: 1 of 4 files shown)


MaochenWang1:
@jukkar @krish2718 @rlubos please review this fix

krish2718 (commented):
Can you please do `git cherr-pick -x <upstream commit>` this way it will add a reference in the commit log with the upstream SHA .

krish2718 (approved):

PR #70, opened 2024-12-18
Feature/RSNO Porting

fengming-ye:
Porting open source wpa_supplicant RSN override commits.


fengming-ye:
Hi @jukkar @krish2718 @rlubos 

could you pls review PR and leave comments?
These are ported from upstream and will be needed for WFA certificate.

MaochenWang1 (approved):

krish2718 (commented):
can we please mark these as `fromtree` and use `git cherry-pick -x` to note the original commit as well? THis would be helpful during upmerge.

krish2718:
> These are ported from upstream

Were there any conflicts? How did you fix the conflicts? Hopefully the original commits haven't been modified?

krish2718:
Also, please raise a manifest update PR with detailed description how these can be used? Do we need to extend the Wi-Fi management layer for this? extend docs/tests?

fengming-ye:
> can we please mark these as `fromtree` and use `git cherry-pick -x` to note the original commit as well? THis would be helpful during upmerge.

updated.



> > These are ported from upstream
> 
> Were there any conflicts? How did you fix the conflicts? Hopefully the original commits haven't been modified?


There are 2 simple conflict caused by zephyr unique change and 1 unimportant conflict which is in wlantest/rx_mgmt.c.
Listed here:
1. fca6d04a2ac046f75f57202abb98b48441dc36e5 surely fixed in wpa.h:366.
2. 4c65cea30edc248734359e76946b1eb121f6df65 surely fixed src/ap/ieee802_11.c:5588
3. 59c608bd8d55089ef8d4aa2f2b7dd46ddd2e4f31 unimportant fixed in wlantest/rx_mgmt.c about kde process. We don't use wlantest.



> Also, please raise a manifest update PR with detailed description how these can be used? Do we need to extend the Wi-Fi management layer for this? extend docs/tests?

The change is transparent to user. So the usage is not changed. I'm raising update manifest PR. And we don't need to extend L2 wifi.

krish2718:
> There are 2 simple conflict caused by zephyr unique change and 1 unimportant conflict which is in wlantest/rx_mgmt.c.

So, those are part of the commit itself? Ideally better to put them in a separate commit to identify those easily.

> The change is transparent to user.

So, how will this feature be used? `wpa_cli` ?

fengming-ye:
> > There are 2 simple conflict caused by zephyr unique change and 1 unimportant conflict which is in wlantest/rx_mgmt.c.
> 
> So, those are part of the commit itself? Ideally better to put them in a separate commit to identify those easily.
> 
> > The change is transparent to user.
> 
> So, how will this feature be used? `wpa_cli` ?


They are just line offset conflict. And these commits will be dropped.

This feature does not touch command process. Just recognise RSN Override IE beffore RSN IE. The other process is the same.

krish2718:
> They are just line offset conflict. And these commits will be dropped when next upmerge.

No, we cannot rewrite history in Zephyr west projects, so, we just do a `git merge` in the next upmerge.

> And I don't know how to put them in separate commit. Did you have example?

If they are small then it's fine, else during cherry-picking we can create new commits for fixing conflicts.

krish2718 (approved):
Please add `CONFIG_TESTING_OPTIONS` to the twister at https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/net/wifi/configs

jukkar (approved):

MaochenWang1:
Hi @jukkar  can you merge this since it's approved?

PR #74, opened 2025-01-02
[noup] zephyr: Create dedicated heap for supplicant

rado17:
Create dedicated heap for supplicant operations. Currently, WPA supplicant uses the k_malloc, which is shared across various subsystems, and we cannot enforce a minimum availability. Using a dedicated HEAP that can be pre-allocated solves this by always having necessary memory.

### DIFF
--- a/src/utils/os_zephyr.c
+++ b/src/utils/os_zephyr.c
@@ -18,6 +18,10 @@ int clock_gettime(clockid_t clock_id, struct timespec *ts);
 #include "includes.h"
 #include "os.h"
 
+#if !defined(CONFIG_WIFI_NM_WPA_SUPPLICANT_GLOBAL_HEAP)
+K_HEAP_DEFINE(wifi_nm_wpa_supplicant_mem_pool, CONFIG_WIFI_NM_WPA_SUPPLICANT_HEAP);
+#endif /* CONFIG_WIFI_NM_WPA_SUPPLICANT_GLOBAL_HEAP */
+
 void os_sleep(os_time_t sec, os_time_t usec)
 {
 	k_sleep(K_USEC(usec + USEC_PER_SEC * sec));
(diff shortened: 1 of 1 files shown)


MaochenWang1:
why we need this dedicated heap for supplicant?
BTW, currently WPA supplicant uses the kernel k_malloc, not the libc malloc

rado17:
@sachinthegreen 

> why we need this dedicated heap for supplicant? BTW, currently WPA supplicant uses the kernel k_malloc, not the libc malloc

Updated the description to correct this. Since the kernel heap is shared across various subsystems, we cannot enforce a minimum availability for supplicant. Using a pre-allocated heap solves this problem.

src/utils/os_zephyr.c
```diff
@@ -18,6 +18,13 @@ int clock_gettime(clockid_t clock_id, struct timespec *ts);
 #include "includes.h"
 #include "os.h"
 
+#if defined(CONFIG_NOCACHE_MEMORY)
+K_HEAP_DEFINE_NOCACHE(wifi_nm_wpa_supplicant_mem_pool, CONFIG_WIFI_NM_WPA_SUPPLICANT_HEAP);
+#else
+K_HEAP_DEFINE(wifi_nm_wpa_supplicant_mem_pool, CONFIG_WIFI_NM_WPA_SUPPLICANT_HEAP);
+#endif /* CONFIG_NOCACHE_MEMORY */
+#define WORD_SIZE 4
```
krish2718, line 26: we don't need `nocache` memory for supplicant.

krish2718 (commented):
Please add a config option to use global or dedicated heap.

src/utils/os_zephyr.c
rado17: Fixed

src/utils/os_zephyr.c
```diff
@@ -209,7 +214,7 @@ char *os_strdup(const char *s)
 }
 
 void *os_memdup(const void *src, size_t len)
-{
+{
```
krish2718, line 217: whitespace?

src/utils/os_zephyr.c
```diff
@@ -220,12 +225,22 @@ void *os_memdup(const void *src, size_t len)
 
 void *os_malloc(size_t size)
 {
+#if defined(CONFIG_WIFI_NM_WPA_SUPPLICANT_HEAP)
+	return k_heap_aligned_alloc(&wifi_nm_wpa_supplicant_mem_pool, WORD_SIZE, size, K_FOREVER);
```
krish2718, line 229: why is this aligned but the common heap isn't? If alignment is needed both version should be aligned, no?

krish2718 (approved):

PR #76, opened 2025-01-24
[noup] zephyr: Update to support MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG

frkv:
-This commit add support for MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
 which adds support for PRNG (seeded by TRNG) without using
 the legacy Mbed TLS APIs in ctr_drbg.c and entropy.c. When the
 configuration MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG is enabled there
 is a new function available in the system that is used to override
 this behavior. This function is called mbedtls_psa_external_get_random
 but for compatibility with the f_rng/p_rng signature the
 function mbedtls_psa_get_random (found in mbedtls/psa_util.h) is
 used directly in code.
-Added a function pointer called hostap_rng_fn which is set
 to mbedtls_psa_get_random if MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG is
 enabled, otherwise it is set to the legacy API mbedtls_ctr_drbg_random.
-Added a context pointer called hostap_rng_ctx that is set to
 MBEDTLS_PSA_RANDOM_STATE (NULL) if MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
 is set, otherwise it points to the existing ctr_drbg context.
-Updated all calls using legacy APIs making use of the f_rng/p_rng
 pattern to use hostap_rng_fn and hostap_rng_ctx in crypto_mbedtls_alt
-Added forward declaration of hostap_rng_fn and hostap_rng_ctx in
 tls_mbedtls_alt.c and changed every API-call making use of the
 f_rng/p_rng pattern.


krish2718:
+ @MaochenWang1 please review

krish2718 (approved):

jukkar (approved):

jukkar:
ping @MaochenWang1

MaochenWang1 (approved):

fengming-ye (approved):

PR #81, opened 2025-03-17
[noup] zephyr: Fix the Wi-Fi secure associations

krish2718:
Due to a recent change the Zephyr network stack sends the protocol in network byte order, so, need to be converted to host order before using it. This was dropping all EAPoL frames and broke secure association.


jukkar (approved):

jerome-pouiller (approved):
> Due to a recent change the Zephyr network stack

If you have the opportunity to mention the commit that introduce the change, it would be great.

krish2718:
> > Due to a recent change the Zephyr network stack
> 
> If you have the opportunity to mention the commit that introduce the change, it would be great.

I did in the manifest PR, if I mention it here github won't render it as a link because it's in zephyr repo.

MaochenWang1 (approved):

PR #87, opened 2025-05-30
[noup] zephyr: crypto: fix coverity issue

Qingling-Wu:
Fix INTEGER_OVERFLOW coverity issue.
crypto_ecdh_set_peerkey:
Check value of the unsigned variable len before decrement. crypto_ec_key_get_subject_public_key/crypto_ec_key_get_ecprivate_key: Check return value of mbedtls_asn1_write_len and mbedtls_asn1_write_tag, if less than zero, return NULL.


jukkar (commented):
The changes are good. There seems to be two separate issues here, could you split the commit to two pieces and resubmit the PR?

jukkar (approved):

MaochenWang1 (approved):

PR #88, opened 2025-06-09
Zephyr: Group cipher addition in beacon

abhinavnxp:
Updated the check for group cipher addition in beacon. Due to earlier check WPA_CIPHER_AES_128_CMAC was not added in beacon.

### DIFF
--- a/src/ap/wpa_auth_ie.c
+++ b/src/ap/wpa_auth_ie.c
@@ -342,7 +342,7 @@ static u8 * rsne_write_data(u8 *buf, size_t len, u8 *pos, int group,
 
 
 	if (mfp != NO_MGMT_FRAME_PROTECTION &&
-	    group_mgmt_cipher != WPA_CIPHER_AES_128_CMAC) {
+	    group_mgmt_cipher != WPA_CIPHER_NONE) {
 		if (2 + 4 > buf + len - pos)
 			return NULL;
 		if (!pmkid) {


src/ap/wpa_auth_ie.c
```diff
@@ -342,7 +342,7 @@ static u8 * rsne_write_data(u8 *buf, size_t len, u8 *pos, int group,
 
 
 	if (mfp != NO_MGMT_FRAME_PROTECTION &&
-	    group_mgmt_cipher != WPA_CIPHER_AES_128_CMAC) {
+	    group_mgmt_cipher != WPA_CIPHER_NONE) {
```
krish2718, line 345: This was done deliberately, see the rationale here https://github.com/zephyrproject-rtos/hostap/commit/44fa5e747b7aca39285e2511d5c94684e0723b6b.patch TL;DR, no need to add default values for:

* PMKIDCount field (2 bytes)
* Group Management Cipher Suite field (4 bytes)

This saves 6 extra bytes in every Beacon and Probe Response frame, even when these fields contained only default values.

So, unless you have a specific case this change isn;t needed, and even if you as Jukka says it needs to be submitted (and preferably merged/accepted) to the root upstream first.

krish2718 (changes requested):
As per above, rejecting this for now.

abhinavnxp:
Closing PR since this change is the expected behavior.

PR #89, opened 2025-06-24
Remove bss max idle period

ajayparida:
Remove bss max idle period related changes from supplicant. To be handled as part of net_mgmt.


krish2718 (approved):

sachinthegreen (approved):

src/utils/os.h
```diff
@@ -381,7 +381,6 @@ int os_strcasecmp(const char *s1, const char *s2);
  */
```
jukkar, line 381: The commit message is confusing, the subject talks about partial revert of one commit, and the commit message body talks about another revert. Is this one now reverting a revert?
Could you reword the commit message and subject? Please put in the subject a short note what this is fixing, and in the commit message more explanation what is being changed.

krish2718:
> @MaochenWang1 please take a look

@MaochenWang1 gentle reminder

PR #90, opened 2025-07-04
[noup] zephyr: Fix build warnings with workaround

D-Triveni:
When a module enables ECDH but does not enable EC, it causes build warnings and twister build failures.
This workaround fixes these issues.

### DIFF
--- a/src/crypto/crypto_mbedtls_alt.c
+++ b/src/crypto/crypto_mbedtls_alt.c
@@ -1750,6 +1750,8 @@ static int crypto_mbedtls_ike_id_from_ecp_group_id(mbedtls_ecp_group_id grp_id)
 #include <mbedtls/ecp.h>
 #include <mbedtls/pk.h>
 
+mbedtls_ecp_keypair *mbedtls_pk_ec(const mbedtls_pk_context pk);
+
 static int crypto_mbedtls_keypair_gen(int group, mbedtls_pk_context *pk)
 {
     mbedtls_ecp_group_id grp_id = crypto_mbedtls_ecp_group_id_from_ike_id(group);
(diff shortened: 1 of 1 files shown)


MaochenWang1:
what‘s the detail of build warnings?
The definition of `mbedtls_pk_ec` is under `MBEDTLS_ECP_C`, why not just enable `MBEDTLS_ECP_C` for your module?

src/crypto/crypto_mbedtls_alt.c
```diff
@@ -1750,6 +1750,8 @@ static int crypto_mbedtls_ike_id_from_ecp_group_id(mbedtls_ecp_group_id grp_id)
 #include <mbedtls/ecp.h>
 #include <mbedtls/pk.h>
 
+mbedtls_ecp_keypair *mbedtls_pk_ec(const mbedtls_pk_context pk);
```
MaochenWang1, line 1753: will this cause no definition of `mbedtls_pk_ec`?

D-Triveni: No, this won't cause a `no definition` error, just a warning related to implicit declaration of the function.

D-Triveni:
> what‘s the detail of build warnings? The definition of `mbedtls_pk_ec` is under `MBEDTLS_ECP_C`, why not just enable `MBEDTLS_ECP_C` for your module?

The build warnings are mostly due to implicit type conversions and incorrect format specifiers.
`mbedtls_pk_ec` is defined under `MBEDTLS_ECP_C`, but enabling it also requires `MBEDTLS_LEGACY_CRYPTO_C`, which brings in legacy crypto APIs not needed for the 54H platform. To keep the build minimal and avoid unnecessary code, we chose not to enable it.

MaochenWang1 (approved):

krish2718:
Please rebase to fix the twister failure.

PR #95, opened 2025-07-17
[noup] l2_packet: Implement a direct function call option

krish2718:
In Zephyr we can leverage using direct function call instead of socket send this gives us:

 * Faster time to send EAPoL OTA
 * Ability to handle M4 vs set_key race
 * Handle rekeying effectively in presence of traffic

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -2519,6 +2519,7 @@ int wpa_drv_hapd_send_eapol(void *priv, const u8 *addr, const u8 *data, size_t d
 #ifdef CONFIG_WIFI_NM_HOSTAPD_AP
 	struct zep_drv_if_ctx *if_ctx = priv;
 	struct hostapd_data *hapd     = NULL;
+	const struct zep_wpa_supp_dev_ops *dev_ops = NULL;
 	int ret                       = -1;
 
 	/* TODO: Unused for now, but might need for rekeying */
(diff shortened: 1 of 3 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -2532,14 +2533,27 @@ int wpa_drv_hapd_send_eapol(void *priv, const u8 *addr, const u8 *data, size_t d
 
 	wpa_printf(MSG_DEBUG, "hostapd: Send EAPOL frame (encrypt=%d)", encrypt);
 
+	/* Try to use driver operation for high-priority transmission */
+	dev_ops = get_dev_ops(if_ctx->dev_ctx);
+#ifdef CONFIG_WIFI_NM_WPA_SUPPLICANT_L2_PKT_DIRECT
```
MaochenWang1, line 2538: Since you add `CONFIG_WIFI_NM_WPA_SUPPLICANT_L2_PKT_DIRECT` related change inside `l2_packet_send`, then all the changes before `l2_packet_send` are not needed, right?

src/drivers/driver_zephyr.h
```diff
@@ -351,6 +351,15 @@ struct zep_wpa_supp_dev_ops {
 	int (*cancel_remain_on_channel)(void *priv);
 	int (*get_inact_sec)(void *if_priv, const u8 *addr);
 	void (*send_action_cancel_wait)(void *priv);
+
+#ifdef CONFIG_NRF70_L2_PACKET
```
MaochenWang1, line 355:
```suggestion
#ifdef CONFIG_WIFI_NM_WPA_SUPPLICANT_L2_PKT_DIRECT
```

src/l2_packet/l2_packet_zephyr.c
```diff
@@ -143,6 +176,21 @@ l2_packet_init(const char *ifname, const u8 *own_addr, unsigned short protocol,
...
 	os_memcpy(l2->own_addr, link_addr->addr, link_addr->len);
 
+	/* Get device and driver operations for high-priority transmission */
+	device = net_if_get_device(l2->iface);
+	if (device) {
+		dev_ops = get_dev_ops(device);
+#ifdef CONFIG_WIFI_NM_WPA_SUPPLICANT_L2_PKT_DIRECT
+		if (dev_ops && dev_ops->send_l2_packet) {
+			/* Get device private data from the interface context */
+			struct nrf_wifi_vif_ctx_zep *vif_ctx_zep = device->data;
```
MaochenWang1, line 186: why use `nrf_wifi_vif_ctx_zep` struct here?

src/l2_packet/l2_packet_zephyr.c
```diff
@@ -41,35 +44,63 @@ int l2_packet_get_own_addr(struct l2_packet_data *l2, u8 *addr)
...
+	}
+
+	/* Get driver operations */
+	dev_ops = get_dev_ops(device);
+	if (!dev_ops) {
+		wpa_printf(MSG_ERROR, "l2_packet_send: Failed to get driver ops");
+		return -1;
+	}
+
+#ifdef CONFIG_WIFI_NM_WPA_SUPPLICANT_L2_PKT_DIRECT
```
sachinthegreen, line 69: Maybe protect the entire set of changes (i.e. starting from line 55) with  `CONFIG_WIFI_NM_WPA_SUPPLICANT_L2_PKT_DIRECT` ? 

(Same comment for changes in other functions too).

sachinthegreen (changes requested):

PR #96, opened 2025-07-25
[noup] hostap: Apply configured keep-alive early schedule offset

ajayparida:
Use the configured value to determine the time offset (in milliseconds) for scheduling Wi-Fi keep-alive frames earlier than the BSS Max Idle Period.

### DIFF
--- a/src/utils/os.h
+++ b/src/utils/os.h
@@ -381,6 +381,7 @@ int os_strcasecmp(const char *s1, const char *s2);
  */
 int os_strncasecmp(const char *s1, const char *s2, size_t n);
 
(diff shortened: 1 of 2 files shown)


src/utils/os.h
```diff
@@ -381,6 +381,8 @@ int os_strcasecmp(const char *s1, const char *s2);
  */
 int os_strncasecmp(const char *s1, const char *s2, size_t n);
 
+#define CONFIG_BSS_MAX_IDLE_TIME CONFIG_WIFI_NM_WPA_SUPPLICANT_BSS_MAX_IDLE_TIME
```
krish2718, line 384: this is unrelated

krish2718 (commented):
Can you please add submit this to usptream first and then fromlist?

ajayparida:
There will be separate PR for this.

PR #98, opened 2025-08-11
[noup] zephyr: Update crypto mbedtls wrapper to support TLSv1.3

nxf58150:
Update tls_mbedtls_alt.c to support TLSv1.3
1. The TLS 1.3 key exporter mechanism is fundamentally different from TLS 1.2.The TLS 1.3 uses the HKDF for key derivation, instead of PRF used in TLS1.2. In tls_connection_export_key function, use HKDF-Expand_Label function if current tls version is TLS1.3.
2. Excluded all tls_prf_type check if TLSv1.3 flag is enabled. This type will not be used in TLSv1.3 handshake process.
3. Add protected result indication handling on EAP-peer side in tls_connection_handshake.

### DIFF
--- a/src/crypto/tls_mbedtls_alt.c
+++ b/src/crypto/tls_mbedtls_alt.c
@@ -58,6 +58,10 @@
 #include <mbedtls/ssl_ticket.h>
 #include <mbedtls/x509.h>
 #include <mbedtls/x509_crt.h>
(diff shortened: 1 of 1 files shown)


MaochenWang1 (approved):

jukkar:
@nxf58150 could you check the conflicts that github is reporting

PR #105, opened 2025-09-25
[noup] zephyr: use standard paths for time.h and signal.h

cfriedt:
Use standard include paths for time.h and signal.h instead of paths prefixed with zephyr/posix/.

### DIFF
--- a/src/utils/os_zephyr.c
+++ b/src/utils/os_zephyr.c
@@ -5,7 +5,8 @@
  * See README for more details.
  */
 
-#include <zephyr/posix/time.h>
+#include <time.h>
+
 #include <zephyr/posix/sys/time.h>
 
 /* The clock_gettime() would be found in <zephyr/posix/time.h> but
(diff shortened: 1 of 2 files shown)


src/utils/os_zephyr.c
```diff
@@ -5,7 +5,8 @@
  * See README for more details.
  */
 
-#include <zephyr/posix/time.h>
+#include <time.h>
+
 #include <zephyr/posix/sys/time.h>
```
krish2718, line 10: can this also be removed now?

cfriedt: In the future, they will likely also include other ISO C headers that are multi-purposes for POSIX too. E.g.

* stdio.h

The nice thing about all of this, is that (by 4.3.0) we will not need to use a `<zephyr/posix/..>` prefix for any POSIX headers, because the build system will know to take those from either the toolchain's libc (for libcs that do support POSIX) or zephyr (for libc's that do not support POSIX).

henrikbrixandersen (approved):

krish2718 (approved):

aescolar (commented):
Note https://github.com/zephyrproject-rtos/zephyr/issues/96566#issuecomment-3335265991

MaochenWang1 (approved):

PR #131, opened 2026-04-01
[noup] zephyr: fix uAP WPS PBC overlap

sanjay-yadav-nxp:
When acting as an AP with multiple stations attempting a WPS session simultaneously, WPS PBC overlap was not detected. This occurred because the driver incorrectly dropped incoming Probe Request frames due to a logic error in the filtering criteria.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -2494,6 +2494,7 @@ static int wpa_drv_zep_set_ap(void *priv,
 	}
 
 	if_ctx->beacon_set = 1;
+	if_ctx->ap_probe_req_listen = true;
 
 	ret = 0;
 
@@ -2587,6 +2588,7 @@ int wpa_drv_zep_stop_ap(void *priv, int link_id)
 out:
 	if (if_ctx) {
 		if_ctx->beacon_set = false;
+		if_ctx->ap_probe_req_listen = false;
 	}
 	return ret;
 }


MaochenWang1:
Hi @sanjay-yadav-nxp 
please change `uAP` to `SAP` in the commit header, create Zephyr manifest PR, and provide more detail about the `logic error`.

krish2718 (approved):

src/drivers/driver_zephyr.c
```diff
@@ -1030,7 +1030,7 @@ static void wpa_drv_zep_event_mgmt_rx(struct zep_drv_if_ctx *if_ctx,
 	fc = le_to_host16(mgmt->frame_control);
 	stype = WLAN_FC_GET_STYPE(fc);
 
-	if (stype == WLAN_FC_STYPE_PROBE_REQ && !if_ctx->probe_req_listen && !if_ctx->ap_probe_req_listen) {
+	if (stype == WLAN_FC_STYPE_PROBE_REQ && (!if_ctx->probe_req_listen || !if_ctx->ap_probe_req_listen)) {
```
jukkar: I have always hard time reading these kind of boolean check where there are multiple `!`, could we change it to
```
if (stype == WLAN_FC_STYPE_PROBE_REQ &&
     !(if_ctx->probe_req_listen && if_ctx->ap_probe_req_listen))
```

MaochenWang1: I am still confused by this logic.
If `ap_probe_req_listen` is true, and `probe_req_listen` is `false`, the old logic, `if (stype == WLAN_FC_STYPE_PROBE_REQ && !if_ctx->probe_req_listen && !if_ctx->ap_probe_req_listen)`, is **`false`**, right? It will not drop the packet.
But the new logic will be **`true`** in this case, and drop the packet.

MaochenWang1 (changes requested):

MaochenWang1 (approved):

jukkar (approved):

PR #136, opened 2026-05-12
bugfix: events: stop processing scan results when aborting

ghost:
In our testing of the Wi-Fi stack, we came across a scenario where when attempting to connect to a non-existing Wi-Fi network and then calling wifi disconnect, the stack gets stuck in an infinite loop where it's continually trying to process the same scan results without end.
This one-line fix prevents that loop from happening by exiting the scan parsing early if we are aborting.

Steps to reproduce the original issue:
```
uart:~$ wifi connect -s nonexistent -p testtesttest -k 10
Connection requested
# Wait about 10 seconds
uart:~$ wifi disconnect
# deadlock
```
Note that this does not always reproduce the issue, it's somewhat timing dependent

Platform: NXP RW610

### DIFF
--- a/wpa_supplicant/events.c
+++ b/wpa_supplicant/events.c
@@ -2487,7 +2487,7 @@ static int _wpa_supplicant_event_scan_results(struct wpa_supplicant *wpa_s,
 	}
 #endif /* CONFIG_NO_RANDOM_POOL */
 
-	if (update_only) {
+	if (update_only || (data && data->scan_info.aborted)) {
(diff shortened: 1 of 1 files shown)


krish2718 (changes requested):
Please add commit log with details of the issue and how/why the fix helps. And also submit this fix to the hostap mailing list, once its merged then it can be cherry-picked to Zephyr as `fromtree`

fabiobaltieri:
User deleted, closing.

PR #144, opened 2026-06-12
rsn: Allow WPA1 TKIP EAPOL-Key under FIPS with PSA MD5

krish2718:
Zephyr hostap builds with CONFIG_FIPS, which strips HMAC-MD5 and RC4 call sites even when the PSA backend provides them. WPA-PSK (WPA1) APs require EAPOL-Key descriptor version 1: HMAC-MD5 for the 4-way MIC and RC4 for group-key unwrap. That breaks association with legacy and QuickTrack WPA-PSK test APs.

When CONFIG_PSA_WANT_ALG_MD5 is set (personal STA crypto backends that explicitly enable legacy TKIP interop), allow the ver=1 MIC and RC4 key data paths. Builds without PSA MD5 keep the existing FIPS blocks.

Assisted-by: Cursor:Auto

### DIFF
--- a/src/common/wpa_common.c
+++ b/src/common/wpa_common.c
@@ -217,11 +217,11 @@ int wpa_eapol_key_mic(const u8 *key, size_t key_len, int akmp, int ver,
 	}
 
 	switch (ver) {
-#ifndef CONFIG_FIPS
+#if !defined(CONFIG_FIPS) || defined(CONFIG_PSA_WANT_ALG_MD5)
 	case WPA_KEY_INFO_TYPE_HMAC_MD5_RC4:
 		wpa_printf(MSG_DEBUG, "WPA: EAPOL-Key MIC using HMAC-MD5");
 		return hmac_md5(key, key_len, buf, len, mic);
-#endif /* CONFIG_FIPS */
+#endif /* !CONFIG_FIPS || CONFIG_PSA_WANT_ALG_MD5 */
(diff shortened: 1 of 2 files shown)


src/common/wpa_common.c
```diff
@@ -217,11 +217,11 @@ int wpa_eapol_key_mic(const u8 *key, size_t key_len, int akmp, int ver,
 	}
 
 	switch (ver) {
-#ifndef CONFIG_FIPS
+#if !defined(CONFIG_FIPS) || defined(CONFIG_PSA_WANT_ALG_MD5)
```
MaochenWang1, line 220: Is it possible that don't define `CONFIG_FIPS` when `CONFIG_PSA_WANT_ALG_MD5` is enabled? as this will not touch these supplicant source code.

krish2718: Yes, the usage of FIPS is here is a hack as MbedTLS is not FIPS compliant, but it helps us disable legacy algos, but as Wi-Fi should always be backwards compatible, we end up adding these exceptions to FIPS.

We could disable these by default and have a non-FIPS build explicitly, I will check and update, thanks for the suggestion.

krish2718:
Closing this as fix is now purely contained in the Zephyr hostap module shim (https://github.com/zephyrproject-rtos/zephyr/pull/111137)

PR #147, opened 2026-07-06
P2p support

nxf58150:
Code enhancement of P2P feature

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -3173,8 +3173,19 @@ int wpa_drv_zep_probe_req_report(void *priv, int report)
 	if_ctx = priv;
 
 	if (!report) {
-		wpa_printf(MSG_DEBUG, "%s: Unregister probe request report", __func__);
+		if (if_ctx->ap_probe_req_listen) {
+			/*
+			 * Do not disable Probe Request reporting while in AP/GO mode,
+			 * it was registered by AP setup independently.
+			 */
+			wpa_printf(MSG_DEBUG,
+				   "%s: Skip disabling probe req report while in AP mode",
+				   __func__);
+			if_ctx->probe_req_listen = false;
+			return 0;
+		}
 		if_ctx->probe_req_listen = false;
+		if_ctx->probe_req_set = false;
 		return 0;
 	}
 
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -2994,9 +2994,7 @@ out:
 int wpa_drv_hapd_send_eapol(void *priv, const u8 *addr, const u8 *data, size_t data_len,
                             int encrypt, const u8 *own_addr, u32 flags, int link_id)
 {
-#ifdef CONFIG_WIFI_NM_HOSTAPD_AP
 	struct zep_drv_if_ctx *if_ctx = priv;
-	struct hostapd_data *hapd     = NULL;
 	int ret                       = -1;
 
 	/* TODO: Unused for now, but might need for rekeying */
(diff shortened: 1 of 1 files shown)


MaochenWang1 (approved):

src/drivers/driver_zephyr.c
```diff
@@ -3179,7 +3166,7 @@ int wpa_drv_zep_probe_req_report(void *priv, int report)
 	int ret = -1;
 
 	if (!priv) {
-		wpa_printf(MSG_ERROR, "%s: Invalid handle", __func__);
+		wpa_printf(MSG_DEBUG, "%s: Unregister probe request report", __func__);
```
jukkar: Shouldn't this change be in the `[noup] zephyr: skip disabling probe req report when AP mode is active` commit?

nxf58150: Updated. Thanks!

src/drivers/driver_zephyr.c
```diff
@@ -3166,15 +3166,26 @@ int wpa_drv_zep_probe_req_report(void *priv, int report)
...
```
gannicus-gj: if priv is NULL, the original log message is good , why need to change

nxf58150: I think this is rebase error. Let me check and update.

src/drivers/driver_zephyr.c
```diff
@@ -3173,8 +3173,20 @@ int wpa_drv_zep_probe_req_report(void *priv, int report)
 	if_ctx = priv;
 
 	if (!report) {
+		if (if_ctx->ap_probe_req_listen) {
```
jukkar: What is the purpose of the 2nd commit?

- **Defensive only**, the guard is there so that a future unregister path cannot
  kill AP/GO reporting. Then keep the guard, drop `probe_req_set = false`, and
  say in the commit message that it is protecting a path that does not exist yet.
- **Actually stop reporting on `report=0`**, then the commit is incomplete: it
  needs an unregister driver op, and only then does clearing `probe_req_set` make
  sense, gated by the AP check exactly as written.

nxf58150: The purpose of the 2nd commit is defensive only. The probe request cannot be unregistered for AP/GO. 
Why drop probe_req_set = false? Once removed, the flag will keep true after setting it to true. For P2P interface, this will cause P2P find fail after P2P GO is teared down.

nxf58150: Commit message updated.

jukkar:
Thanks for the answer, that helps. The failure you describe sounds real, but I do
not think `probe_req_report()` is where it should be fixed.

I asked LLM review as it is difficult to follow the control flow manually. It had these comments:

The `!report` path of `wpa_drv_zep_probe_req_report()` makes no driver call at all, on any driver. It is
host side bookkeeping and a `return 0`. So it cannot be the moment at which the
driver loses the Probe Request registration. What makes your patch work is that
`wpas_start_listen()` and `wpas_stop_listen()` call `probe_req_report(1)` and `(0)`
once per listen cycle (`p2p_supplicant.c:2716`, `:2798`), so the flag happens to get
cleared shortly before the next `p2p_find`. That also means the "already registered"
early return at `:3194` never fires again on the P2P path, and every listen cycle now
costs an extra `register_frame()` round trip to the firmware.

Zephyr cannot use a separate P2P group interface
(`wpa_driver_zep_ops` has no `.if_add`, so `wpas_p2p_create_iface()` always returns
0), so the GO and the P2P device role share one `if_ctx`. Leaving GO mode is what
invalidates the registration: on the one in-tree driver that implements these ops,
`nrf_wifi_wpa_supp_deinit_ap()` ends in
`nrf_wifi_iftype_change(vif_ctx_zep, NRF_WIFI_IFTYPE_STATION)` when `vif_idx == 0`,
which takes the vif down and issues a change-vif. A per-vif frame registration does
not survive that.

Which driver did you reproduce and verify this on, and does it
implement the `deinit_ap` op? I ask because the in-tree NXP driver
(`nxp_wifi_drv_ops`, `drivers/wifi/nxp/src/nxp_wifi_drv.c:2577`) implements neither
`register_frame` nor `register_mgmt_frame` nor `deinit_ap`, and that combination has
an awkward consequence for this patch:

GO teardown goes `wpa_supplicant_ap_deinit()` -> `wpa_drv_deinit_ap()`
(`wpa_supplicant/ap.c:1207`) -> `wpa_drv_zep_deinit_ap()`, and `hostapd_drv_stop_ap()`
is not on that path. `wpa_drv_zep_deinit_ap()` returns early at
`if (!dev_ops || !dev_ops->deinit_ap)`, which is *before*
`if_ctx->ap_probe_req_listen = false`. So on a driver without `deinit_ap`,
`ap_probe_req_listen` goes true at the first GO and never comes back down, every
later `probe_req_report(0)` takes the new AP active branch, and the
`probe_req_set = false` line you added is unreachable.

If your driver does implement `deinit_ap`, great, and then `deinit_ap` is precisely
where the registration is known to be gone. If it does not, the patch cannot be doing
what it looks like it is doing, and that is worth sorting out first. Either way the
answer points at the AP mode transition rather than at `report(0)`.

**Suggested shape.** Clear the flag where the registration is actually lost, in the
`out:` blocks so it still runs when the driver does not implement the op, next to the
`beacon_set` reset that is already handled that way:

```c
 out:
      if (if_ctx) {
              if_ctx->beacon_set = false;
              if_ctx->ap_probe_req_listen = false;
+             if_ctx->probe_req_set = false;
      }
```

in wpa_drv_zep_stop_ap(), and the same in wpa_drv_zep_deinit_ap(), which also
moves the existing ap_probe_req_listen = false out of the success only path. That
is precise, it costs no re-registration per listen cycle, and it also covers the case
where the GO is started and torn down without an intervening listen stop. Assuming
the registration is lost is the safe direction: being wrong costs one redundant
register_frame(), while a stale true costs a p2p_find that never sees a Probe
Request.

With that in place the AP guard in probe_req_report() has nothing left to protect,
since both branches reduce to probe_req_listen = false; return 0; and differ only
in the log message. So commit 2 would become the one or two lines above, plus a
commit message describing the actual failure: P2P find stops working after a GO is
torn down, because the driver drops the Probe Request registration when the interface
leaves AP mode. That is worth writing down properly, since [noup] patches get
re-applied by hand on every hostap rebase and the commit body is the only record.

Two small things on the current message: "doesn not exit yet" should be "does not
exist yet", and it still says the probe_req_set line is there "to keep state
consistent", which is not the reason you gave here.

jukkar (approved):

PR #149, opened 2026-07-13
[noup] hostap: Add WPA-PSK-SHA384 and FT-PSK-SHA384 key management

kapbh:
Introduce WPA_KEY_MGMT_PSK_SHA384 and WPA_KEY_MGMT_FT_PSK_SHA384 and wire them through config parsing, RSN AKM mapping, and RSN IE generation for SHA-384 based PSK and fast-transition PSK suites.

Assisted-by: Claude:Auto


krish2718:
Can you clarify the use case before this goes further? A couple of concerns:

**1. Justification / `[noup]` rationale.** The commit message describes *what* the patch does but not *why* — which feature, product, or AP interop scenario requires `WPA-PSK-SHA384` / `FT-PSK-SHA384`? These are AKM suites `00-0F-AC:20` and `:19`. Upstream deliberately reserved the selector constants back in 2022 (`5456b0f26`, "Define new RSN AKM suite selector values", from IEEE P802.11-REVme/D1.3) but never wired them into key management — while implementing the siblings from that same commit that had real drivers (802.1X-SHA384 `:23`, SAE-EXT-KEY `:24`/`:25`). The reason there's no upstream implementation is that no Wi-Fi Alliance program uses PSK with SHA-384: WPA3-Personal is SAE, and the 192-bit/CNSA suite is EAP/802.1X-SHA384, not PSK. If upstream has no use case, we should have a concrete one documented in the commit before carrying a `[noup]`. What is it?

**2. The patch is functionally incomplete — it will not actually work.** It wires up AKM *negotiation* (config parse/write, RSN AKM ↔ bit mapping, RSN IE emission) but not the *cryptography*. Specifically:
- `wpa_key_mgmt_sha384()` in `src/common/defs.h` is not updated to include the two new bits (nor `wpa_key_mgmt_sha256()`).
- As a result, in `wpa_pmk_to_ptk()` these AKMs match neither the SHA-384 nor SHA-256 branch and fall through to the final `else`, deriving the PTK with `sha1_prf()`. The same `wpa_key_mgmt_sha384(akmp)` gate guards ~8 other sites (EAPOL-Key MIC, key-data encryption, PMKID/KDK derivation), all of which will take the wrong branch.
- The key-length tables `wpa_kck_len()` / `wpa_kek_len()` / `wpa_mic_len()` fall to their `default:` (16/16/16) instead of the SHA-384 sizes (24-byte KCK/MIC, 32-byte KEK).

Since Zephyr uses the software crypto path (MbedTLS, no offload), this code is actually exercised — a 4-way handshake with these AKMs would derive keys with SHA-1 and wrong lengths and fail MIC verification. So even for its stated purpose the patch is non-functional as written.

Given both points, I'd hold this until (a) the use case is documented in the commit message, and (b) if it's genuinely needed, the crypto side is completed to match. Happy to help scope (b) if the use case checks out.

MaochenWang1:
This non-zephyr specific change which touches the supplicant core code, it should be push to supplicant upstream first

PR #150, opened 2026-07-17
Fix CVE 2025 24912

MarkoSagadin:
Hello!

In one of our projects that uses the Nordic's nRF7002 chip and by extension the WiFi stack we have detected a vulnerability in this hostap repository with the [FOSSA](https://fossa.com/) tool.

The screenshot show below comes from the tool:

<details><summary>Screenshot</summary>
<p>

image

</p>
</details> 

The tool reports that our version of the hostap tool contains the CVE-2025-24912. More about this specific CVE can be found here: https://nvd.nist.gov/vuln/detail/CVE-2025-24912.

The NIST website references two commits that fix this issue:

* https://git.w1.fi/cgit/hostap/commit/?id=726432d7622cc0088ac353d073b59628b590ea44
* https://git.w1.fi/cgit/hostap/commit/?id=339a334551ca911187cc870f4f97ef08e11db109

I have checked our version of hostap, as well as the latest main, but it seems that this fork branched of from the upstream, before the fix commits were applied.

This PR contains those two commits, I cherry-picked them from the upstream.

What I would like to know:
1. Can this be merged into the Zephyr's fork of hostap?
2. Do the cherry-picked commits, applied to the fork, fix the same bug that was discovered in the upstream? They applied cleanly, however I don't have sufficient knowledge to confirm that.

### DIFF
--- a/src/radius/radius_client.c
+++ b/src/radius/radius_client.c
@@ -1259,13 +1259,6 @@ static void radius_client_receive(int sock, void *eloop_ctx, void *sock_ctx)
 		       roundtrip / 100, roundtrip % 100);
 	rconf->round_trip_time = roundtrip;
 
-	/* Remove ACKed RADIUS packet from retransmit list */
-	if (prev_req)
-		prev_req->next = req->next;
-	else
-		radius->msgs = req->next;
-	radius->num_msgs--;
-
 	for (i = 0; i < num_handlers; i++) {
 		RadiusRxResult res;
 		res = handlers[i].handler(msg, req->msg, req->shared_secret,
@@ -1276,6 +1269,13 @@ static void radius_client_receive(int sock, void *eloop_ctx, void *sock_ctx)
 			radius_msg_free(msg);
 			/* fall through */
 		case RADIUS_RX_QUEUED:
+			/* Remove ACKed RADIUS packet from retransmit list */
+			if (prev_req)
+				prev_req->next = req->next;
+			else
+				radius->msgs = req->next;
+			radius->num_msgs--;
+
 			radius_client_msg_free(req);
 			return;
 		case RADIUS_RX_INVALID_AUTHENTICATOR:
@@ -1297,7 +1297,6 @@ static void radius_client_receive(int sock, void *eloop_ctx, void *sock_ctx)
 		       msg_type, hdr->code, hdr->identifier,
 		       invalid_authenticator ? " [INVALID AUTHENTICATOR]" :
 		       "");
-	radius_client_msg_free(req);
 
  fail:
 	radius_msg_free(msg);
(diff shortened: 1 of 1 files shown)


krish2718:
Heads-up on scope: `radius_client.c` (the file both fixes touch) is not compiled in the Zephyr hostap build at all.

In `zephyr/modules/hostap/CMakeLists.txt` the AP build defines `CONFIG_NO_RADIUS` unconditionally, and none of `src/radius/*.c` (`radius.c`, `radius_client.c`, `radius_das.c`, `radius_server.c`) are added to `zephyr_library_sources`. So there is no RADIUS client object in any Zephyr image.

How hostapd + enterprise works on Zephyr without a RADIUS client:

- Zephyr's hostapd runs as an integrated EAP authentication server (equivalent to `eap_server=1`), not as a RADIUS authenticator that relays EAP to an external AAA server.
- The 802.1X RADIUS relay path is guarded by `#ifndef CONFIG_NO_RADIUS` (`ieee802_1x_aaa_send` -> `ieee802_1x_encapsulate_radius` in `src/ap/ieee802_1x.c`), so it is a no-op. EAP is terminated locally via `eapol_auth_sm.c`, `authsrv.c`, `eap_server*.c`, and the local user DB (`eap_user_db.c`).
- The enterprise Kconfig only enables the `eap_server_*` methods (TLS/TTLS/PEAP/GTC/MSCHAPV2); none of them link the RADIUS client.

Implication for CVE-2025-24912: the vulnerable path (pending-request handling on receiving a RADIUS Access response in `radius_client.c`) is unreachable in Zephyr, since Zephyr's hostapd is never a RADIUS client and the file is not built. So this is not a live/exploitable vulnerability in the current configuration.

On the fix itself: hostap is planning a 2.12 release and Zephyr will do an upmerge onto it, so both of these commits will land through that upmerge anyway. I'd prefer we take them that way rather than a manual cherry-pick here. If we do decide to cherry-pick in the meantime, it should be a clean two-commit cherry-pick preserving both upstream commits as-is (not squashed/reworked).

krish2718 (changes requested):
Requesting changes per my comment above: the affected file (radius_client.c) is compiled out in the Zephyr hostap build (CONFIG_NO_RADIUS), so CVE-2025-24912 is not reachable here. These fixes will come in naturally via the upcoming hostap 2.12 upmerge, which I'd prefer over a manual cherry-pick. If we do cherry-pick in the meantime, it must be a clean two-commit cherry-pick preserving both upstream commits as-is.

MarkoSagadin:
Hello @krish2718, thank you for the immediate answer!

Ok, it is good to know that we aren't affected by this, since the affected files are compiled out and that the enterprise WPA is implemented without RADIUS client.

I agree with you, it is best that the fixes come into this repo with the upmerge.

Thus, I will close this PR.

PR #151, opened 2026-07-20
eap_server: rename local state enum to avoid STM32 HAL collision

zoltangyarmati:
stm32n6xx.h defines ErrorStatus as a global-scope enum with a SUCCESS member. The anonymous enum inside struct eap_identity_data puts its CONTINUE/SUCCESS/FAILURE members at file scope too, causing a redeclaration error in C17 when building on STM32 targets.

Rename to EAP_ID_{CONTINUE,SUCCESS,FAILURE} to avoid the conflict.

### DIFF
--- a/src/eap_server/eap_server_identity.c
+++ b/src/eap_server/eap_server_identity.c
@@ -13,7 +13,7 @@
 
 
 struct eap_identity_data {
-	enum { CONTINUE, SUCCESS, FAILURE } state;
+	enum { EAP_ID_CONTINUE, EAP_ID_SUCCESS, EAP_ID_FAILURE } state;
 	int pick_up;
 };
 
@@ -25,7 +25,7 @@ static void * eap_identity_init(struct eap_sm *sm)
 	data = os_zalloc(sizeof(*data));
 	if (data == NULL)
 		return NULL;
-	data->state = CONTINUE;
+	data->state = EAP_ID_CONTINUE;
 
 	return data;
 }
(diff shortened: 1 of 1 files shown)


src/eap_server/eap_server_identity.c
```diff
@@ -13,7 +13,7 @@
 
 
 struct eap_identity_data {
-	enum { CONTINUE, SUCCESS, FAILURE } state;
+	enum { EAP_ID_CONTINUE, EAP_ID_SUCCESS, EAP_ID_FAILURE } state;
```
MaochenWang1, line 16: Thanks for the PR, but modifying supplicant core code is not the right approach here.

This file is upstream hostap code from w1.fi. Carrying a local rename creates a permanent diff and ongoing merge burden on every upstream sync.
The root cause is STM32 HAL polluting the global namespace with generic names like SUCCESS. Preferred alternatives:
Fix include isolation at the build system level so HAL headers don't leak into hostap compilation units.
Use a platform shim that #undefs the conflicting symbols.

zoltangyarmati: Dear MaochenWang1,

thanks for the reply. I assume then that your first sentence was supposed to be
"Thanks for the PR, but modifying supplicant core code is _NOT_ the right approach here."
OK, gonna ditch this PR and figure out an other approach.

MaochenWang1: Sorry, I typed it incorrectly, updated

